PR #11, opened 2026-08-09
fix(security): eliminate ReDoS vulnerability in filter_results

numikel:
## Summary

This PR eliminates the **Regular Expression Denial of Service (ReDoS)** vulnerability in `filter_results` where attacker-supplied patterns like `(.+)+!` could permanently freeze the server.

## The Problem

**Before this fix:**
- `filter_results(pattern="(.+)+!", ...)` freezes the server for ~40 seconds and makes `/health` unreachable
- Attack vector: any caller can provide a regex pattern, no authentication required
- Root cause: synchronous `re.compile()` with catastrophic backtracking on the event loop

**After this fix:**
- Patterns are compiled with **RE2** (guarantees linear-time complexity)
- Same pattern now executes in <1ms
- Unsupported syntax (lookaround, backreferences) is rejected before execution with a validation error

## Changes

**8 commits implementing the fix:**

1. **Dependency**: Add `google-re2>=1.1.20251105`
2. **New module**: `services/pattern_matching.py` — single place to safely compile client-supplied patterns with RE2
3. **Configuration**: New settings `LAW_MCP_MAX_PATTERN_LENGTH` (default 512, range 64–4096, clamped with startup warning) and `LAW_MCP_FILTER_MAX_RECORDS` (default 100)
4. **Core fix**: Replace `re.compile()` with `re2.compile()` in `ResultStore.filter_results()`; cap result set size with fail-fast error
5. **Startup**: Inject configuration limits into services; warn if pattern length is clamped
6. **Cleanup**: Remove dead exception handler in `DocumentStore.search()`
7. **Documentation**: Update `filter_results` description to list supported regex subset and record limit
8. **History**: Record changes in CHANGELOG.md

## Test Results

✅ **Unit tests**: 247 passed (was 190, +57 new tests for all 8 tasks)
✅ **Integration tests**: 41 passed
✅ **Quality**: `ruff check`, `ruff format --check`, `mypy` all pass

Regression test `test_catastrophic_pattern_returns_promptly` has a 5-second timeout and fails on the unpatched code.

## Behavioral Changes

- Patterns longer than `LAW_MCP_MAX_PATTERN_LENGTH` (default 512, configurable) are now rejected before compilation
- Result sets larger than `LAW_MCP_FILTER_MAX_RECORDS` (default 100, configurable) now return an error instead of being processed
  - This ensures "no matches found" always means "the entire set was searched" (not just a partial subset)
- Patterns with lookaround `(?=...)`, `(?<=...)`, `(?!...)` or backreferences `\1` are rejected with a validation error

## For Reviewers

- **Security**: Verify `services/pattern_matching.py` uses RE2 API correctly, `max_mem` limit is set, errors are translated to Polish
- **Functional**: Check result-set cap enforcement in `result_store.py` and config-injection in `server.py`
- **Tests**: All new tests in `test_pattern_matching.py`, `test_result_store.py`, `test_config.py`, and `test_server.py` pass

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,7 @@ docker-compose.override.yml
 .ruff_cache/
 .mypy_cache/
 .pytest_cache/
+
+# Other
+docs/
+.worktrees/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - **`filter_results` breaking change: result set size cap** — Previously unlimited, calls on result sets larger than 100 records (`LAW_MCP_FILTER_MAX_RECORDS`) are now refused, even without a `pattern`; the response reports `error_category: precondition` so it's clear the request needs adjusting. Narrow the search or raise `LAW_MCP_FILTER_MAX_RECORDS` to work around it.
-- **`filter_results`: pattern compilation has bounded resource budgets** — Some simple-looking patterns, e.g. `\p{L}{0,112}`, are now rejected because their compiled size exceeds the limit. Patterns with more than four variable range quantifiers (such as `{1,900}`) are also rejected before compilation to avoid blocking the event loop.
+- **`filter_results`: pattern compilation has bounded resource budgets** — Some simple-looking patterns, e.g. `\p{L}{0,112}`, are now rejected because their compiled size exceeds the limit. Patterns with more than four variable range quantifiers (such as `{1,900}`) are also rejected before compilation to avoid blocking the event loop. The preflight recognizes only exact RE2 POSIX tokens (`[[:name:]]` / `[[:^name:]]`); incomplete `[.` / `[=` forms cannot hide ranges from the guard.
 - **`filter_results` parameter description names the supported pattern syntax** — The `pattern` parameter description now documents the supported regex subset, matching the validation error message.
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,13 +18,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - **`filter_results` breaking change: result set size cap** — Previously unlimited, calls on result sets larger than 100 records (`LAW_MCP_FILTER_MAX_RECORDS`) are now refused, even without a `pattern`; the response reports `error_category: precondition` so it's clear the request needs adjusting. Narrow the search or raise `LAW_MCP_FILTER_MAX_RECORDS` to work around it.
-- **`filter_results`: pattern compilation has a bounded compiled-size budget** — Some simple-looking patterns, e.g. `\p{L}{0,112}`, are now rejected because their compiled size exceeds the limit. Matching is linear-time and compiled program size is capped; compilation *time* is not separately bounded.
+- **`filter_results`: pattern compilation has bounded resource budgets** — Some simple-looking patterns, e.g. `\p{L}{0,112}`, are now rejected because their compiled size exceeds the limit. Patterns with more than four variable range quantifiers (such as `{1,900}`) are also rejected before compilation to avoid blocking the event loop.
 - **`filter_results` parameter description names the supported pattern syntax** — The `pattern` parameter description now documents the supported regex subset, matching the validation error message.
 
 ### Added
 
 - **New dependency: `google-re2`** — Ships as prebuilt binary wheels for CPython 3.13 on Windows and manylinux; no compiler needed to install.
 - **`LAW_MCP_MAX_PATTERN_LENGTH`** (default 512, range 64-4096) — Out-of-range values are clamped with a startup warning instead of blocking the server.
+- **`LAW_MCP_FILTER_MAX_RECORDS`** (default 100, floor 1) — Limits records processed by a single `filter_results` call; operators may raise it, but very high values lengthen the synchronous scan.
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,31 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Security
+
+- **`filter_results`: fixed ReDoS vulnerability** — Client-supplied patterns now compile through a linear-time matching engine instead of Python's backtracking `re`; lookaround and backreferences are rejected with a validation error instead of executed. Previously, `pattern="(.+)+!"` could freeze the server indefinitely.
+
+### Fixed
+
+- **HTTP client resource leak on shutdown** — If the server exited through an exception or interruption while running, the HTTP client was never closed; shutdown now always releases it.
+
+### Changed
+
+- **`filter_results` breaking change: result set size cap** — Previously unlimited, calls on result sets larger than 100 records (`LAW_MCP_FILTER_MAX_RECORDS`) are now refused, even without a `pattern`; the response reports `error_category: precondition` so it's clear the request needs adjusting. Narrow the search or raise `LAW_MCP_FILTER_MAX_RECORDS` to work around it.
+- **`filter_results`: pattern compilation has a bounded compiled-size budget** — Some simple-looking patterns, e.g. `\p{L}{0,112}`, are now rejected because their compiled size exceeds the limit. Matching is linear-time and compiled program size is capped; compilation *time* is not separately bounded.
+- **`filter_results` parameter description names the supported pattern syntax** — The `pattern` parameter description now documents the supported regex subset, matching the validation error message.
+
+### Added
+
+- **New dependency: `google-re2`** — Ships as prebuilt binary wheels for CPython 3.13 on Windows and manylinux; no compiler needed to install.
+- **`LAW_MCP_MAX_PATTERN_LENGTH`** (default 512, range 64-4096) — Out-of-range values are clamped with a startup warning instead of blocking the server.
+
+### Removed
+
+- **Dead exception handler in document search** — Removed an unreachable `try/except` around escaped-query compilation.
+
 ## [2.4.0] - 2026-07-08
 
 ### Security
--- a/README.md
+++ b/README.md
@@ -179,7 +179,7 @@ All settings are configured via environment variables with the `LAW_MCP_` prefix
 | `LAW_MCP_CIRCUIT_BREAKER_RECOVERY_TIMEOUT` | `60.0` | Seconds before trying recovery |
 | `LAW_MCP_CIRCUIT_BREAKER_HALF_OPEN_MAX_CALLS` | `3` | Test calls in half-open state |
 | `LAW_MCP_MAX_PATTERN_LENGTH` | `512` | Max `filter_results` pattern length, clamped to 64-4096 |
-| `LAW_MCP_FILTER_MAX_RECORDS` | `100` | Max records `filter_results` processes per call, floor 1 |
+| `LAW_MCP_FILTER_MAX_RECORDS` | `100` | Max records `filter_results` processes per call; floor 1, no ceiling (very high values lengthen the synchronous but linear scan) |
 | `LAW_MCP_LOG_LEVEL` | `INFO` | Log level: DEBUG, INFO, WARNING, ERROR |
 | `LAW_MCP_LOG_FORMAT` | `text` | Log format: `text` or `json` |
 
--- a/README.md
+++ b/README.md
@@ -178,6 +178,8 @@ All settings are configured via environment variables with the `LAW_MCP_` prefix
 | `LAW_MCP_CIRCUIT_BREAKER_THRESHOLD` | `5` | Failures before circuit breaker opens |
 | `LAW_MCP_CIRCUIT_BREAKER_RECOVERY_TIMEOUT` | `60.0` | Seconds before trying recovery |
 | `LAW_MCP_CIRCUIT_BREAKER_HALF_OPEN_MAX_CALLS` | `3` | Test calls in half-open state |
+| `LAW_MCP_MAX_PATTERN_LENGTH` | `512` | Max `filter_results` pattern length, clamped to 64-4096 |
+| `LAW_MCP_FILTER_MAX_RECORDS` | `100` | Max records `filter_results` processes per call, floor 1 |
 | `LAW_MCP_LOG_LEVEL` | `INFO` | Log level: DEBUG, INFO, WARNING, ERROR |
 | `LAW_MCP_LOG_FORMAT` | `text` | Log format: `text` or `json` |
 
--- a/src/law_scrapper_mcp/config.py
+++ b/src/law_scrapper_mcp/config.py
@@ -46,7 +46,7 @@ class Settings(BaseSettings):
 
     @property
     def effective_max_pattern_length(self) -> int:
-        """Return max pattern length clamped to the allowed range (D3.1)."""
+        """Return max pattern length clamped to the allowed range"""
         return min(
             max(self.max_pattern_length, MAX_PATTERN_LENGTH_FLOOR),
             MAX_PATTERN_LENGTH_CEILING,
@@ -82,7 +82,7 @@ class Settings(BaseSettings):
 
 
 def log_pattern_limit_clamping(current: Settings, log: logging.Logger) -> None:
-    """Log a warning when the max pattern length setting was clamped (D3.1)."""
+    """Log a warning when the max pattern length setting was clamped"""
     if not current.max_pattern_length_was_clamped:
         return
     log.warning(
--- a/src/law_scrapper_mcp/config.py
+++ b/src/law_scrapper_mcp/config.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import logging
 from typing import Literal
 
 from pydantic_settings import BaseSettings, SettingsConfigDict
@@ -10,6 +11,12 @@ from pydantic_settings import BaseSettings, SettingsConfigDict
 # do najbliższej granicy, a nie odrzucana — start serwera nie może się nie udać.
 MAX_PATTERN_LENGTH_FLOOR = 64
 MAX_PATTERN_LENGTH_CEILING = 4096
+
+# Dolna granica liczby rekordów, jakie filter_results może zwrócić. Wartość 0 (lub ujemna)
+# oznacza, że KAŻDE wywołanie na niepustym zestawie wyników skończy się odmową —
+# narzędzie stałoby się trwale niesprawne. Górnej granicy celowo nie ma: to pozostaje
+# w gestii operatora, podobnie jak reszta pól tego pliku (bez Field()).
+FILTER_MAX_RECORDS_FLOOR = 1
 
 
 class Settings(BaseSettings):
@@ -40,6 +47,35 @@ class Settings(BaseSettings):
     doc_store_max_size_bytes: int = 5 * 1024 * 1024
     doc_store_ttl: int = 7200
 
+    # Filtrowanie wyników (filter_results)
+    # Parametry operacyjne (D3): strojalne przez operatora. Nie są zabezpieczeniem
+    # przed ReDoS — tę rolę pełni silnik RE2 w services/pattern_matching.py.
+    max_pattern_length: int = 512
+    filter_max_records: int = 100
+
+    @property
+    def effective_max_pattern_length(self) -> int:
+        """Limit długości wzorca po przycięciu do widełek (D3.1)."""
+        return min(
+            max(self.max_pattern_length, MAX_PATTERN_LENGTH_FLOOR),
+            MAX_PATTERN_LENGTH_CEILING,
+        )
+
+    @property
+    def max_pattern_length_was_clamped(self) -> bool:
+        """Czy skonfigurowana wartość musiała zostać przycięta do widełek."""
+        return self.effective_max_pattern_length != self.max_pattern_length
+
+    @property
+    def effective_filter_max_records(self) -> int:
+        """Limit rekordów po przycięciu do dolnej granicy.
+
+        Asymetrycznie wobec `max_pattern_length` — bez osobnego ostrzeżenia w logu:
+        wartość efektywna jest i tak obserwowalna w komunikacie ResultSetTooLargeError
+        przy pierwszym wywołaniu filter_results, więc cichy floor nie ukrywa problemu.
+        """
+        return max(self.filter_max_records, FILTER_MAX_RECORDS_FLOOR)
+
     # Circuit breaker
     circuit_breaker_threshold: int = 5
     circuit_breaker_recovery_timeout: float = 60.0
@@ -52,6 +88,19 @@ class Settings(BaseSettings):
     # Server info
     server_name: str = "law-scrapper-mcp"
     server_version: str = "2.4.0"
+
+
+def log_pattern_limit_clamping(current: Settings, log: logging.Logger) -> None:
+    """Zgłoś w logu fakt przycięcia limitu długości wzorca (D3.1, punkt 1)."""
+    if not current.max_pattern_length_was_clamped:
+        return
+    log.warning(
+        "LAW_MCP_MAX_PATTERN_LENGTH=%d jest poza dozwolonym zakresem %d-%d; serwer używa wartości efektywnej %d.",
+        current.max_pattern_length,
+        MAX_PATTERN_LENGTH_FLOOR,
+        MAX_PATTERN_LENGTH_CEILING,
+        current.effective_max_pattern_length,
+    )
 
 
 settings = Settings()
--- a/src/law_scrapper_mcp/config.py
+++ b/src/law_scrapper_mcp/config.py
@@ -7,15 +7,8 @@ from typing import Literal
 
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
-# Widełki limitu długości wzorca (D3.1). Wartość spoza zakresu jest przycinana
-# do najbliższej granicy, a nie odrzucana — start serwera nie może się nie udać.
 MAX_PATTERN_LENGTH_FLOOR = 64
 MAX_PATTERN_LENGTH_CEILING = 4096
-
-# Dolna granica liczby rekordów, jakie filter_results może zwrócić. Wartość 0 (lub ujemna)
-# oznacza, że KAŻDE wywołanie na niepustym zestawie wyników skończy się odmową —
-# narzędzie stałoby się trwale niesprawne. Górnej granicy celowo nie ma: to pozostaje
-# w gestii operatora, podobnie jak reszta pól tego pliku (bez Field()).
 FILTER_MAX_RECORDS_FLOOR = 1
 
 
@@ -47,15 +40,13 @@ class Settings(BaseSettings):
     doc_store_max_size_bytes: int = 5 * 1024 * 1024
     doc_store_ttl: int = 7200
 
-    # Filtrowanie wyników (filter_results)
-    # Parametry operacyjne (D3): strojalne przez operatora. Nie są zabezpieczeniem
-    # przed ReDoS — tę rolę pełni silnik RE2 w services/pattern_matching.py.
+    # Filtering
     max_pattern_length: int = 512
     filter_max_records: int = 100
 
     @property
     def effective_max_pattern_length(self) -> int:
-        """Limit długości wzorca po przycięciu do widełek (D3.1)."""
+        """Return max pattern length clamped to the allowed range (D3.1)."""
         return min(
             max(self.max_pattern_length, MAX_PATTERN_LENGTH_FLOOR),
             MAX_PATTERN_LENGTH_CEILING,
@@ -63,16 +54,16 @@ class Settings(BaseSettings):
 
     @property
     def max_pattern_length_was_clamped(self) -> bool:
-        """Czy skonfigurowana wartość musiała zostać przycięta do widełek."""
+        """Return whether the configured max pattern length was clamped."""
         return self.effective_max_pattern_length != self.max_pattern_length
 
     @property
     def effective_filter_max_records(self) -> int:
-        """Limit rekordów po przycięciu do dolnej granicy.
+        """Return filter max records raised to the minimum allowed value.
 
-        Asymetrycznie wobec `max_pattern_length` — bez osobnego ostrzeżenia w logu:
-        wartość efektywna jest i tak obserwowalna w komunikacie ResultSetTooLargeError
-        przy pierwszym wywołaniu filter_results, więc cichy floor nie ukrywa problemu.
+        Asymmetric vs `max_pattern_length` — no separate log warning: the
+        effective value is already visible in the ResultSetTooLargeError message
+        on the first filter_results call, so a silent floor does not hide the issue.
         """
         return max(self.filter_max_records, FILTER_MAX_RECORDS_FLOOR)
 
@@ -91,7 +82,7 @@ class Settings(BaseSettings):
 
 
 def log_pattern_limit_clamping(current: Settings, log: logging.Logger) -> None:
-    """Zgłoś w logu fakt przycięcia limitu długości wzorca (D3.1, punkt 1)."""
+    """Log a warning when the max pattern length setting was clamped (D3.1)."""
     if not current.max_pattern_length_was_clamped:
         return
     log.warning(
--- a/src/law_scrapper_mcp/config.py
+++ b/src/law_scrapper_mcp/config.py
@@ -6,6 +6,11 @@ from typing import Literal
 
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
+# Widełki limitu długości wzorca (D3.1). Wartość spoza zakresu jest przycinana
+# do najbliższej granicy, a nie odrzucana — start serwera nie może się nie udać.
+MAX_PATTERN_LENGTH_FLOOR = 64
+MAX_PATTERN_LENGTH_CEILING = 4096
+
 
 class Settings(BaseSettings):
     """Application settings with environment variable support."""
--- a/src/law_scrapper_mcp/server.py
+++ b/src/law_scrapper_mcp/server.py
@@ -11,7 +11,7 @@ from starlette.responses import JSONResponse
 from law_scrapper_mcp.client.cache import TTLCache
 from law_scrapper_mcp.client.circuit_breaker import CircuitBreaker
 from law_scrapper_mcp.client.sejm_client import SejmApiClient
-from law_scrapper_mcp.config import settings
+from law_scrapper_mcp.config import log_pattern_limit_clamping, settings
 from law_scrapper_mcp.logging_config import setup_logging
 from law_scrapper_mcp.services.act_service import ActService
 from law_scrapper_mcp.services.changes_service import ChangesService
@@ -29,6 +29,7 @@ logger = logging.getLogger(__name__)
 async def lifespan(server):
     """Initialize and cleanup server resources."""
     logger.info("Starting Law Scrapper MCP Server v%s", settings.server_version)
+    log_pattern_limit_clamping(settings, logger)
 
     circuit_breaker = CircuitBreaker(
         failure_threshold=settings.circuit_breaker_threshold,
@@ -51,27 +52,32 @@ async def lifespan(server):
     )
     content_processor = ContentProcessor()
 
-    result_store = ResultStore()
+    result_store = ResultStore(
+        max_pattern_length=settings.effective_max_pattern_length,
+        pattern_length_limit_clamped=settings.max_pattern_length_was_clamped,
+        max_records=settings.effective_filter_max_records,
+    )
     metadata_service = MetadataService(client)
     search_service = SearchService(client)
     act_service = ActService(client, document_store, content_processor)
     changes_service = ChangesService(client)
 
-    yield {
-        "client": client,
-        "cache": cache,
-        "document_store": document_store,
-        "content_processor": content_processor,
-        "result_store": result_store,
-        "metadata_service": metadata_service,
-        "search_service": search_service,
-        "act_service": act_service,
-        "changes_service": changes_service,
-    }
-
-    await client.close()
-    await cache.clear()
-    logger.info("Law Scrapper MCP Server stopped")
+    try:
+        yield {
+            "client": client,
+            "cache": cache,
+            "document_store": document_store,
+            "content_processor": content_processor,
+            "result_store": result_store,
+            "metadata_service": metadata_service,
+            "search_service": search_service,
+            "act_service": act_service,
+            "changes_service": changes_service,
+        }
+    finally:
+        await client.close()
+        await cache.clear()
+        logger.info("Law Scrapper MCP Server stopped")
 
 
 app = FastMCP(
--- a/src/law_scrapper_mcp/services/document_store.py
+++ b/src/law_scrapper_mcp/services/document_store.py
@@ -95,7 +95,13 @@ class DocumentStore:
             return None
 
     async def search(self, eli: str, query: str, context_chars: int = 500) -> list[SearchHit]:
-        """Search within a loaded document."""
+        """Search literal text within a loaded document.
+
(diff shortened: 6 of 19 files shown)


src/law_scrapper_mcp/services/pattern_matching.py
```diff
@@ -0,0 +1,127 @@
...
+            outside the RE2-supported subset, exceeds the compilation memory
+            budget (`RE2_MAX_MEM_BYTES`), or contains characters that cannot be
+            encoded as UTF-8 — e.g. a lone UTF-16 surrogate (U1).
+    """
+    if len(pattern) > max_length:
+        raise PatternValidationError(_too_long_message(len(pattern), max_length, limit_was_clamped))
+
+    try:
+        return re2.compile(pattern, _build_options())
+    except (re2.error, UnicodeError) as e:
```
numikel, line 264: **Critical — compile-time ReDoS survives the engine migration**

RE2 guarantees linear-time *matching*, but it says nothing about *compilation*. Concatenated bounded repetitions expand during simplification, and the cost grows super-linearly in the number of concatenated terms. I reproduced this against this exact function, in a fresh process, with `max_length=512`:

| pattern | length | compile time | outcome |
|---|---|---|---|
| `"a{1,900}" * 5` | 40 | 44.5 ms | accepted |
| `"a{1,900}" * 20` | 160 | 871 ms | accepted |
| `"a{1,900}" * 40` | 320 | 2 716 ms | accepted |
| `"a{1,900}" * 62` | **496** | **6 563 ms** | **accepted** |

Why this matters as much as the bug being fixed:

- The pattern is 496 characters, i.e. **inside the default `max_pattern_length=512`** — no configuration change required.
- It uses nothing but `{n,m}`, which the `pattern` parameter description explicitly advertises as supported. An attacker needs no knowledge of the implementation, only the tool's own documentation.
- `RE2_MAX_MEM_BYTES` does **not** catch it. Every row above was *accepted*, not rejected — the compiled program fits the 2 MiB budget comfortably. `max_mem` bounds the compiled size, so it cannot bound the time spent getting there.
- The result-set cap does not help either: it is evaluated before compilation, and a single stored record passes it. One `search_legal_acts` plus one `filter_results` is still the whole chain, still unauthenticated.

A second reviewer measured 19.6 s for the same shape on the same machine; I measured 6.5 s. The spread does not matter much — both are far outside anything an event loop should absorb.

One caveat on the obvious fix: moving this call to `run_in_executor` or `asyncio.to_thread` lets the *coroutine* time out, but the worker thread keeps burning CPU while holding the GIL through native RE2 code, which reproduces the "freeze survives client disconnect" property of the original bug. A bound that actually holds needs either a pre-flight structural check on the pattern (reject concatenated or nested range quantifiers above an empirically calibrated threshold, before `re2.compile` is reached) or compilation in a separate process with a hard timeout and `terminate()`.

The CHANGELOG entry already says compilation time is not separately bounded, so this is a known limitation — but at six to twenty seconds on default settings it deserves to be tracked as a continuation of F01 rather than as a documented footnote.

src/law_scrapper_mcp/services/result_store.py
```diff
@@ -131,13 +145,13 @@ async def filter_results(
...
 
-        # Regex pattern filter
+        # Pattern filter — RE2 engine, linear-time matching
         if pattern is not None:
-            try:
-                compiled = re.compile(pattern, re.IGNORECASE)
-            except re.error as e:
-                raise ValueError(f"Invalid regex pattern: {e}") from e
-
+            compiled = compile_pattern(
```
numikel, line 150: **Critical (same defect, call-site half)** — this is a synchronous call inside `async def filter`, with no `run_in_executor` or `to_thread` anywhere in this module (grep confirms `compile_pattern` at L150 is the only reference). So the multi-second compilation measured in the comment on `pattern_matching.py` blocks the entire event loop, including `/health` in the same process.

Note also the ordering: the `max_records` guard at L131-132 runs *before* this line, which is correct for the scan cost but means it provides no protection here at all — one stored record is enough to reach compilation.

src/law_scrapper_mcp/services/pattern_matching.py
```diff
@@ -0,0 +1,127 @@
...
+from typing import Any
+
+import re2
+
+from law_scrapper_mcp.config import (
+    MAX_PATTERN_LENGTH_CEILING,
+    MAX_PATTERN_LENGTH_FLOOR,
+)
+
+RE2_MAX_MEM_BYTES = 2 * 1024 * 1024
```
numikel, line 26: **Important — deliberate divergence from the spec, recorded nowhere**

`RE2_MAX_MEM_BYTES = 2 MiB` departs from decision D3, which specified 8 MB on the grounds that it is "the RE2 default".

The divergence itself is right, and I would keep it. A value equal to the library default constrains nothing — `re2.Options().max_mem` is exactly `8388608` — so a security-class constant set to the default would be decorative. And `google-re2` caches compiled patterns in a module-global `functools.lru_cache(typed=True)` (`re2/__init__.py:149`), so the budget multiplies by the cache size: at 8 MiB the ceiling on memory held by *legitimate* client patterns is roughly 1 GiB, at 2 MiB roughly 256 MiB. That is a memory-exhaustion vector independent of ReDoS and absent from the original analysis.

The problem is that none of this is written down anywhere under version control. The spec still says 8 MB, the code says 2 MiB, and `docs/` is gitignored so the spec is not even part of this PR. The next maintainer will reconcile the code to the spec in good faith and restore the ~1 GiB ceiling with no signal that they are reversing a deliberate decision.

The boundary value was also established by measurement — at 1 MiB the compilable limit for `\\p{L}{0,n}` sat around n=55, rejecting innocuous patterns like `\\p{L}{0,200}(?:zdrow|apteka)\\p{L}{0,200}`; at 2 MiB it is n=111. Without that number any future adjustment is guesswork.

Suggested fix: put the rationale in the module docstring, which is versioned and describes the module's contract rather than an implementation detail.

numikel (commented):
## Holistic review against the cluster-1 spec and implementation plan

Scope: full `main...klaster-1` diff (18 files, +985/-40), assessed against the design doc and implementation plan for cluster 1 — task integration, architecture, regressions, security, test quality, merge readiness.

### Quality gates — all exit 0 at `499af4d`

```
ruff check src/ tests/            All checks passed!                            exit=0
ruff format --check src/ tests/  56 files already formatted                    exit=0
mypy src/law_scrapper_mcp/       Success: no issues found in 38 source files   exit=0
pytest tests/unit/               247 passed in 39.59s                          exit=0
pytest tests/integration/        41 passed                                     exit=0
```

### What this PR gets right

The backtracking vector is genuinely closed, not merely mitigated. Every client-supplied pattern flows through `compile_pattern`, both `except` branches raise `PatternValidationError`, and there is no fallback to `re` anywhere — so D2 holds in the way that actually matters, rather than pushing the attacker into a weaker branch. Configuration is wired end to end (`Settings` -> lifespan -> `ResultStore` -> `compile_pattern`) and the *effective*, not raw, values reach the constructor. The `try/finally` around the lifespan `yield` is a real resource-leak fix that went beyond the cluster's scope and was recorded honestly as its own CHANGELOG entry.

### One finding blocks the "F01 is closed" claim

**The migration to RE2 removes search-time backtracking but leaves a compile-time blow-up that produces a comparable outcome.** A pattern built only from the `{n,m}` syntax this tool advertises, sized inside the default `max_pattern_length=512`, takes multiple seconds to compile — synchronously, on the event loop. Details and measurements are in the inline comment on `pattern_matching.py`. The CHANGELOG already states that compilation time is not separately bounded, so this is a known limitation rather than an oversight; my point is that the measured magnitude puts it in the same risk class as the original BLOCKER, not in the footnote class.

I would still merge this — it is a clear net improvement and reverting it would be wrong. But the original audit gated the HTTP transport on F01, and on the evidence below that gate should stay shut until the compile path is bounded.

### A note above the code

`.gitignore:64` ignores all of `docs/`, so **neither the spec, nor the implementation plan, nor the audit report is in the repository**. The acceptance protocol for this PR — 14 acceptance criteria and decisions D1-D7 — exists only on the author's machine. A reviewer on GitHub cannot see the document the PR is being judged against, and changes to design decisions are neither versioned nor reviewable. Worth considering a `!docs/superpowers/` exception, or moving these documents outside `docs/`.

### What I did not verify

I did not reproduce the freeze end to end through a running HTTP server with concurrent `/health` probes, and I did not check `uv.lock` wheel coverage for `cp313-win_amd64` / `manylinux` (acceptance criterion 13). The compile-time finding was reproduced directly against the production `compile_pattern` in a fresh process; the event-loop impact is inferred from the call site being synchronous inside `async def`.

src/law_scrapper_mcp/services/pattern_matching.py
```diff
@@ -0,0 +1,127 @@
...
+    options.max_mem = RE2_MAX_MEM_BYTES
+    return options
+
+
+def _is_pattern_too_large(e: re2.error) -> bool:
+    """Detect a compilation memory-budget overrun (`RE2_MAX_MEM_BYTES`).
+
+    RE2 signals this with the same exception type as a syntax error (`re2.error`),
+    so the message text must be inspected to avoid suggesting a syntax fix when
+    the only problem is pattern complexit.
```
numikel: nit: typo in the docstring — `the only problem is pattern complexit.` should be `complexity.`

tests/unit/test_services/test_pattern_matching.py
```diff
@@ -0,0 +1,193 @@
...
+        chosen empirically JUST ABOVE the threshold (n=14/98 chars still
+        compiles, n=15/105 does not). The calibration is intentionally tight:
+        it guards the `RE2_MAX_MEM_BYTES` constant itself — if someone raised
+        it back to 8 MiB, this pattern would compile again and the test would
+        fail loudly instead of silently stopping to check anything.
+        The message must talk about complexity, not syntax, and must not
+        include `SUPPORTED_SYNTAX_HINT` — otherwise the client gets instructions
+        to remove lookarounds/backreferences the pattern does not contain.
+        """
+        pattern = ".{1000}" * 15
```
numikel, line 255: **Important — the complexity tests are calibrated on the safe shape**

`".{1000}" * 15` uses *exact* repetition, and that shape has a sharp, benign threshold: it is rejected by `max_mem` almost immediately, so this test passes for the right reason but proves less than it appears to.

The expensive shape is *range* repetition (`{min,max}` with min != max) concatenated many times — `"a{1,900}" * 62` is accepted and takes ~6.5 s to compile at 496 characters. No test in this file or in `test_result_store.py` exercises it, so the whole suite passes in ~40 s while leaving the actual vector uncovered.

Suggested fix: add a case with a concatenated range quantifier near `max_pattern_length`, asserting under a deliberately low `@pytest.mark.timeout` (1 s). Without it, a future change to the compile path could regress here silently.

src/law_scrapper_mcp/services/document_store.py
```diff
@@ -101,10 +101,7 @@ async def search(self, eli: str, query: str, context_chars: int = 500) -> list[S
             doc.last_accessed = time.time()
 
             hits = []
-            try:
-                pattern = re.compile(re.escape(query), re.IGNORECASE)
-            except re.error:
-                pattern = re.compile(re.escape(query), re.IGNORECASE)
+            pattern = re.compile(re.escape(query), re.IGNORECASE)
```
numikel, line 110: **Minor — acceptance criterion 8 is formally unmet**

`re.escape` is the only thing preventing client-supplied `query` from becoming a raw pattern under the backtracking `re` engine, i.e. from reproducing F01 here right after it was closed in `filter_results`.

The behaviour is correct and the tests guard it. But finding F51 was never really about the dead `try/except`: it was about the construct *looking* as though a raw regex with a fallback were permitted here. D6 and acceptance criterion 8 required that removing the dead code leave behind a statement of what removing the escaping would cost. That statement went away in `dbe3f6a`, so criterion 8 is unmet even though the first half of the requirement was carried out.

Two acceptable routes: record it in the `search()` docstring, or amend D6 and criterion 8 in the spec so the acceptance protocol does not assert something that was deliberately not implemented. Leaving the mismatch silent is the worst of the three, because criterion 8 will be ticked off unchecked at the next review.

src/law_scrapper_mcp/tools/error_handling.py
```diff
@@ -24,6 +24,7 @@
     InvalidEliError: "validation",
     DocumentNotLoadedError: "precondition",
     ResultSetNotFoundError: "precondition",
+    ResultSetTooLargeError: "precondition",
```
numikel, line 27: **Minor — load-bearing invariant with nothing holding it in place**

`_classify_error` (L37) iterates this dict in insertion order and returns on the *first* matching `isinstance`, so ordering is part of the contract rather than cosmetic.

This change is directly affected: `PatternValidationError` subclasses `ValueError`, and `ValueError` sits lower in this dict. Classification is correct today. Reordering the entries — alphabetising them, or adding a new exception "somewhere near the top" — would silently change the error category returned to the MCP client. Neither mypy nor the current tests would catch it, since the types stay valid and only the category value changes.

Suggested fix: state the invariant in the `_classify_error` docstring, where anyone editing the dict will see it.

src/law_scrapper_mcp/services/result_store.py
```diff
@@ -189,7 +222,7 @@ def __init__(self, result_set_id: str):
 _SORTABLE_FIELDS = {"title", "eli", "year", "pos", "status", "type", "promulgation_date", "effective_date"}
 
 
-def _match_field(act: ActSummaryOutput, field: str, compiled: re.Pattern[str]) -> bool:
+def _match_field(act: ActSummaryOutput, field: str, compiled: CompiledPattern) -> bool:
```
numikel, line 225: info (out of cluster scope, flagged so it does not get lost): `_match_field` silently falls back to `title` for an unrecognised field name (L228) instead of raising. This is the same class of problem as the "silent degradation to literal matching" that D2 explicitly rejected — the client gets results computed over a different field than requested, with no signal. The spec places it out of scope, so it is not a merge blocker.

src/law_scrapper_mcp/config.py
```diff
@@ -2,10 +2,15 @@
...
 from __future__ import annotations
 
+import logging
 from typing import Literal
 
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
+MAX_PATTERN_LENGTH_FLOOR = 64
+MAX_PATTERN_LENGTH_CEILING = 4096
+FILTER_MAX_RECORDS_FLOOR = 1
```
numikel, line 12: info: `FILTER_MAX_RECORDS_FLOOR = 1` has no upper counterpart. That is consistent with D3 — this is an operational parameter, so tunability is intended, and the floor correctly prevents a 0/negative value from making the tool permanently inoperable. Worth a line in the README noting that a very large value restores a long synchronous scan on the event loop: linear thanks to RE2, but long.

CHANGELOG.md
```diff
@@ -5,6 +5,31 @@ All notable changes to this project will be documented in this file.
...
+
+- **`filter_results` breaking change: result set size cap** — Previously unlimited, calls on result sets larger than 100 records (`LAW_MCP_FILTER_MAX_RECORDS`) are now refused, even without a `pattern`; the response reports `error_category: precondition` so it's clear the request needs adjusting. Narrow the search or raise `LAW_MCP_FILTER_MAX_RECORDS` to work around it.
+- **`filter_results`: pattern compilation has a bounded compiled-size budget** — Some simple-looking patterns, e.g. `\p{L}{0,112}`, are now rejected because their compiled size exceeds the limit. Matching is linear-time and compiled program size is capped; compilation *time* is not separately bounded.
+- **`filter_results` parameter description names the supported pattern syntax** — The `pattern` parameter description now documents the supported regex subset, matching the validation error message.
+
+### Added
+
+- **New dependency: `google-re2`** — Ships as prebuilt binary wheels for CPython 3.13 on Windows and manylinux; no compiler needed to install.
+- **`LAW_MCP_MAX_PATTERN_LENGTH`** (default 512, range 64-4096) — Out-of-range values are clamped with a startup warning instead of blocking the server.
```
numikel, line 27: nit: `LAW_MCP_FILTER_MAX_RECORDS` appears only inside the *Changed* prose at L20. A reader scanning *Added* for new environment variables will see `LAW_MCP_MAX_PATTERN_LENGTH` and miss the second one.

Separately, this CHANGELOG is strong — it names the breaking change plainly, gives the concrete rejected pattern `\\p{L}{0,112}` instead of a generality, and states outright that compilation time is not separately bounded. That last sentence would have been easy to omit; my only suggestion is to give it more weight, since the measured cost of that unbounded path is several seconds on default settings (see the comment on `pattern_matching.py`).

README.md
```diff
@@ -178,6 +178,8 @@ All settings are configured via environment variables with the `LAW_MCP_` prefix
 | `LAW_MCP_CIRCUIT_BREAKER_THRESHOLD` | `5` | Failures before circuit breaker opens |
 | `LAW_MCP_CIRCUIT_BREAKER_RECOVERY_TIMEOUT` | `60.0` | Seconds before trying recovery |
 | `LAW_MCP_CIRCUIT_BREAKER_HALF_OPEN_MAX_CALLS` | `3` | Test calls in half-open state |
+| `LAW_MCP_MAX_PATTERN_LENGTH` | `512` | Max `filter_results` pattern length, clamped to 64-4096 |
+| `LAW_MCP_FILTER_MAX_RECORDS` | `100` | Max records `filter_results` processes per call, floor 1 |
```
numikel: nit: this row says `floor 1` but does not say there is no ceiling, unlike the row above where the `64-4096` range is closed at both ends. A reader may reasonably assume symmetry.

src/law_scrapper_mcp/server.py
```diff
@@ -51,27 +52,32 @@ async def lifespan(server):
...
-        "metadata_service": metadata_service,
-        "search_service": search_service,
-        "act_service": act_service,
-        "changes_service": changes_service,
-    }
-
-    await client.close()
-    await cache.clear()
-    logger.info("Law Scrapper MCP Server stopped")
+    try:
```
numikel, line 65: Worth calling out positively: wrapping `yield` in `try/finally` is a real resource-leak fix — before this change, exiting the server through an exception skipped `client.close()` and `cache.clear()`. It was outside the cluster's scope and was recorded honestly as its own *Fixed* entry in the CHANGELOG rather than folded quietly into a ReDoS commit.

src/law_scrapper_mcp/services/pattern_matching.py
numikel: Fixed locally in 8579762. An O(n) preflight scanner now rejects more than four variable {min,max} quantifiers before re2.compile; the regression test covers a{1,900} * 62 and requires it to finish in under 0.5 s. The branch is still one commit ahead locally.

src/law_scrapper_mcp/services/result_store.py
numikel: Fixed by 8579762. The expensive shape is rejected before re2.compile, so this synchronous call site no longer blocks the event loop.

src/law_scrapper_mcp/services/pattern_matching.py
numikel: Added the rationale for the 2 MiB limit and the calibration of the four-variable-range limit to the module docstring in 8579762.

src/law_scrapper_mcp/services/pattern_matching.py
numikel: Corrected the docstring typo in 8579762.

tests/unit/test_services/test_pattern_matching.py
numikel: Added a regression test for a{1,900} * 62. Before the fix it compiled for several seconds without raising; now validation completes before compilation.

src/law_scrapper_mcp/services/document_store.py
numikel: Added the contract to the DocumentStore.search docstring: the query remains escaped before use with Python re, and removing that escaping would reintroduce the ReDoS risk.

src/law_scrapper_mcp/tools/error_handling.py
numikel: The _classify_error documentation now records the ordering invariant: subclasses must precede base classes because the first isinstance match determines the category.

src/law_scrapper_mcp/services/result_store.py
numikel: Left unchanged: this is existing behavior outside this patch scope, not a regression introduced by the RE2 migration. Changing it needs a separate public error contract for an invalid field.

src/law_scrapper_mcp/config.py
numikel: README now explicitly documents the absence of an upper limit and the operational consequence: the scan remains linear, but a very high value makes it synchronously longer.

CHANGELOG.md
numikel: CHANGELOG now lists LAW_MCP_FILTER_MAX_RECORDS under Added. The compilation entry also describes the new preflight limit for variable range quantifiers.

README.md
numikel: The README row now states: floor 1, no ceiling, with a warning that high values lengthen the synchronous scan.

src/law_scrapper_mcp/services/pattern_matching.py
```diff
@@ -0,0 +1,186 @@
...
+    classes prevents literals such as `\{1,900}` and `[{1,900}]` from being
+    counted as quantifiers.
+    """
+    quantifier_count = 0
+    in_character_class = False
+    index = 0
+
+    while index < len(pattern):
+        character = pattern[index]
+        if character == "\\":
```
numikel, line 169: **Critical — `\Q...\E` bypass keeps the compile-time DoS open**

RE2 treats `\Q...\E` as quoted literal text. With `pattern = r"\Q[\E" + "a{1,900}" * 62` (501 characters, below the default 512 limit), this scanner skips `\Q`, interprets the quoted literal `[` as a real class opener, and then ignores all 62 variable ranges.

Against current HEAD (`84766bd`), `_has_too_many_variable_range_quantifiers()` returned `False` and the production `compile_pattern()` took **16.45 s** synchronously. The official RE2 syntax reference confirms that `\Q...\E` is accepted. Because this call still runs on the event loop, the unauthenticated compile-time DoS remains reachable.

The guard needs a single-pass tokenizer for the explicitly supported subset and must reject every construct it does not model, including `\Q...\E`, before calling RE2. Add this exact 501-character pattern as a subprocess-isolated timeout regression.

src/law_scrapper_mcp/services/pattern_matching.py
```diff
@@ -0,0 +1,186 @@
...
+            index += 2
+            continue
+        if character == "[" and not in_character_class:
+            in_character_class = True
+        elif character == "]" and in_character_class:
```
numikel: **Important — valid documented POSIX-class syntax is falsely rejected**

The scanner closes character-class state at the inner `]` of a POSIX class. I verified that `"[[:alpha:]{1,2}]" * 5` compiles under raw RE2, while `compile_pattern()` rejects it for supposedly containing five variable range quantifiers. Every brace here is literal class content.

POSIX classes such as `[[:alpha:]]` are explicitly documented as supported, so this introduces a compatibility break beyond the intended four-quantifier policy. Model POSIX class tokens inside the outer character class and add positive cases combining POSIX classes, escaped brackets, and literal braces.

src/law_scrapper_mcp/services/pattern_matching.py
```diff
@@ -0,0 +1,186 @@
...
+            in_character_class = False
+        elif character == "{" and not in_character_class:
+            closing_brace = pattern.find("}", index + 1)
```
numikel: **Minor — the preflight is not strictly linear**

Each unmatched `{` calls `pattern.find("}", index + 1)`, rescanning the remaining suffix. A pattern made of unmatched opening braces therefore has quadratic worst-case behavior. The 4096-character hard cap keeps this from being an independent blocker today, but it contradicts the O(n) contract recorded for this guard.

Keep a forward-only cursor or tokenize braces in the same single pass used to address the lexical-state issues above.

tests/unit/test_services/test_pattern_matching.py
```diff
@@ -0,0 +1,207 @@
...
+        assert "4096" in message
+        assert "przycięcia" in message
+
+    def test_length_error_omits_clamping_note_when_not_clamped(self) -> None:
+        with pytest.raises(PatternValidationError) as exc_info:
+            compile_pattern("a" * 600, max_length=512)
+
+        assert "przycięcia" not in str(exc_info.value)
+
+    def test_many_bounded_ranges_are_rejected_before_compilation(self) -> None:
```
numikel: **Important — the regression proves only the originally reported spelling**

This covers 62 plain ranges, and the positive case covers four small ranges, but there is no first-rejected boundary test for five and no lexical-adversary coverage. A threshold accidentally changed from 4 to 61 would still pass both tests. The current suite also misses `\Q...\E`, POSIX classes containing literal braces, escaped brackets/braces, and malformed delimiters.

Add four/five boundary tests and table-driven tokenizer cases. Potentially blocking compilation regressions should run in a killable subprocess; `pytest-timeout` cannot reliably interrupt native code that holds the GIL.

.gitignore
```diff
@@ -59,3 +59,7 @@ docker-compose.override.yml
 .ruff_cache/
 .mypy_cache/
 .pytest_cache/
+
+# Other
+docs/
```
numikel, line 64: **Important — the security design and acceptance protocol are excluded from history**

This PR newly ignores all of `docs/`. `git ls-files` confirms that both the cluster-1 specification and implementation plan used to judge this PR are not tracked. The local spec also records custom pattern heuristics as rejected, while `8579762` introduces a lexical heuristic with a security-critical threshold.

The code docstring preserves only a condensed rationale; it does not version the accepted grammar, threat model, benchmark method, or criteria 1–14. Track these documents or a focused ADR and update the decision record for the compile-time threat and scanner policy before merge.

numikel (commented):
## Second-round review against the cluster-1 spec and plan

Scope: full `main...klaster-1` at `84766bd`, with focus on the compile-time bound introduced in `8579762` and the documentation cleanup in `84766bd`. Checked against the local cluster-1 specification and implementation plan (D1–D7, acceptance criteria 1–14), the prior review thread, and independent probes against the production `compile_pattern` path.

### Quality gates at `84766bd`

```
ruff check src/ tests/            All checks passed!
ruff format --check src/ tests/  56 files already formatted
mypy src/law_scrapper_mcp/       Success: no issues found in 38 source files
pytest tests/unit/               249 passed
pytest tests/integration/        41 passed
```

`google-re2==1.1.20251105` remains current; both `cp313-win_amd64` and `manylinux` wheels are published.

### What this revision gets right

- Matching-time ReDoS (`(.+)+!`) is closed: client patterns go through RE2 with no `re` fallback.
- The exact reported compile-time shape (`"a{1,900}" * 62`) is rejected before `re2.compile`.
- The 2 MiB `max_mem`, case-insensitive matching, and silenced absl stderr logging are wired correctly.
- Polish validation/precondition errors reach the MCP client surface.
- Pattern-length clamping and the record cap are injected into `ResultStore` end to end.
- Prior documentation findings (`DocumentStore.re.escape` contract, `_classify_error` ordering, CHANGELOG/README notes for `FILTER_MAX_RECORDS`) remain in place.

### Findings that still block merge

#### Critical

1. **`\Q...\E` bypass keeps the compile-time DoS open** — see inline comment on `pattern_matching.py`. Against current HEAD the scanner returned `False` for `r"\Q[\E" + "a{1,900}" * 62` (501 chars, under the default 512 limit) and production `compile_pattern()` took **16.45 s** synchronously on the event loop. This leaves the earlier critical finding only partially resolved.

#### Important

1. **Valid documented POSIX-class syntax is falsely rejected** — see inline on the class-state transition.
2. **Security tests prove only the originally reported spelling** — missing five-range boundary and lexical-adversary coverage.
3. **The acceptance protocol remains unversioned** — `.gitignore` newly ignores all of `docs/`, so the specification and plan used to judge this PR are not in repository history.

#### Minor

1. **The preflight is not strictly linear** — unmatched `{` rescans via `find()`, giving quadratic worst-case behavior under the 4096-character hard cap.

### Re-verification of prior findings

| Prior finding | Status |
|---|---|
| Compile-time blow-up from concatenated variable ranges | Partially resolved — exact spelling blocked; `\Q...\E` bypass still reaches multi-second compilation |
| Synchronous compilation blocks the event loop | Unresolved — the bypass still executes synchronous `re2.compile()` for >10 s |
| 2 MiB rationale not versioned | Partially resolved — condensed rationale in module docstring; detailed D3.2 / scanner policy remain untracked |
| Docstring typo | Resolved |
| Missing expensive variable-range regression | Resolved for the exact reported shape; broader lexical coverage still missing |
| Missing `DocumentStore.re.escape` contract | Resolved |
| Missing `_classify_error` ordering invariant | Resolved |
| Unknown-field fallback to `title` | Intentionally unchanged / out of scope |
| No ceiling for `FILTER_MAX_RECORDS` | Intentionally unchanged and documented |
| CHANGELOG / README notes for the new env vars | Resolved |

### Acceptance criteria

Verified from tests/docs on this checkout: 3, 4, 5, 6, 7, 8, 9, 11, 12, 14.  
Not fully verified end-to-end here: 1 (`/health` under live load), 2 (exact documented patterns at result level), 10 (red run against unpatched head), 13 (clean reinstall / container build — wheels themselves confirmed).  
Security goal still fails because the quoted-literal bypass permits blocking compilation.

### Assessment

**Ready to merge?** No.

The latest scanner blocks one spelling of the attack but is bypassable by valid RE2 syntax within the default pattern limit, leaving synchronous compile-time event-loop DoS exploitable. Fix the tokenizer (including `\Q...\E` and POSIX classes), harden the regression suite, and version the security decision record before merging.

Note: GitHub rejects `REQUEST_CHANGES` on an author-owned PR, so this round is submitted as `COMMENT` with the same merge-blocking verdict.

.gitignore
numikel: Purposeful action. We are not making any changes.

src/law_scrapper_mcp/services/pattern_matching.py
numikel: Fixed in ec334fa. The tokenizer now recognizes POSIX character-class tokens inside bracket expressions, so literal braces remain class content. Positive coverage includes POSIX classes plus escaped brackets and braces.

.gitignore
numikel: The `docs/` policy is intentional and has been handled separately, so this security-scanner follow-up does not change it.

src/law_scrapper_mcp/services/pattern_matching.py
numikel: Fixed in ec334fa. The preflight is now a forward-only tokenizer that rejects `\Q` and `\E` before RE2 compilation. A subprocess-isolated regression covers the exact 501-character bypass and fails after one second if compilation escapes the guard.

tests/unit/test_services/test_pattern_matching.py
numikel: Fixed in ec334fa. Tests now assert the four/five boundary, quoted-literal rejection, POSIX and escaped literal cases, and the exact bypass in a killable subprocess with a one-second timeout.

src/law_scrapper_mcp/services/pattern_matching.py
numikel: Fixed in ec334fa. `{min,max}` recognition now uses a forward-only cursor rather than `find()`, so unmatched opening braces do not rescan suffixes. The regression exercises 4,096 unmatched braces.

src/law_scrapper_mcp/services/pattern_matching.py
```diff
@@ -0,0 +1,227 @@
...
+    while index < len(pattern):
+        character = pattern[index]
+        if character == "\\":
+            if index + 1 < len(pattern) and pattern[index + 1] in {"Q", "E"}:
+                raise PatternValidationError(_quoted_literal_message())
+            index += 2
+            continue
+
+        if in_character_class:
+            if character == "[" and index + 1 < len(pattern) and pattern[index + 1] in {":", ".", "="}:
```
numikel: **Critical - `[.` / `[=` nested-token handling creates another compile-time DoS bypass**

The scanner treats `[. ... .]` and `[= ... =]` inside a bracket expression as nested tokens and scans forward to their closing delimiter. RE2's supported POSIX form that needs special handling here is `[[:name:]]`; the broader handling lets an ordinary class prefix hide every later range quantifier from the preflight.

I reproduced this against current Head with:

`"[a[.b]" + "a{1,900}" * 62 + ".]"`

The pattern is 504 characters, so it passes the default 512-character limit. `_validate_variable_range_quantifier_limit()` skips all 62 ranges, and production `compile_pattern()` exceeded an independent three-second hard timeout in a killable subprocess. The analogous `[=` spelling has the same scanner state transition. Because compilation remains synchronous in `ResultStore.filter_results`, this restores the unauthenticated event-loop DoS; `max_mem` does not bound time already spent simplifying the pattern.

Recognize only RE2-supported POSIX classes - preferably an exact `[[:name:]]` whitelist - or conservatively reject ambiguous nested bracket constructs before RE2. Add subprocess-isolated regressions for both `[.` and `[=` spellings.

tests/unit/test_services/test_pattern_matching.py
```diff
@@ -0,0 +1,274 @@
...
```
numikel: **Important - the expensive-range regression still runs in the pytest process**

`compile_pattern("a{1,900}" * 62, ...)` executes directly in the pytest worker. If the preflight is removed or bypassed, native RE2 compilation blocks before the elapsed-time assertion can run, so the test designed to catch this regression can itself monopolize CI.

Route this case, and every known expensive compilation adversary, through `_compile_in_subprocess` with a hard timeout. Keep direct in-process tests only for inputs proven to complete quickly regardless of guard behavior.

tests/unit/test_services/test_pattern_matching.py
```diff
@@ -0,0 +1,274 @@
...
+import re2
+
+from law_scrapper_mcp.services.pattern_matching import (
+    SUPPORTED_SYNTAX_HINT,
+    PatternValidationError,
+    compile_pattern,
+)
+
+CATASTROPHIC_PATTERN = "(.+)+!"
+COMPILE_TIMEOUT_SECONDS = 1
```
numikel: **Minor - the one-second subprocess timeout is too close to Windows process-startup cost**

The timeout includes interpreter startup and imports, not only validation. It produced a `TimeoutExpired` once during this review; immediate reruns completed near the boundary. This makes a security regression test susceptible to scheduler-load flakes.

Increase the timeout while keeping it well below the demonstrated multi-second attack duration, or start a worker process first and time only the compile request.

numikel (commented):
## Third-round review against the cluster-1 spec and plan

Scope: full `main...klaster-1` from `c444584` through `ec334fa`, with focus on the forward-only tokenizer and regression coverage added after the second round. Checked against the local cluster-1 specification and implementation plan (D1-D7, acceptance criteria 1-14), the prior review threads, current RE2 syntax documentation, and independent subprocess probes against the production `compile_pattern` path.

### Quality gates at ec334fa

```
ruff check src/ tests/            All checks passed!
ruff format --check src/ tests/  56 files already formatted
mypy src/law_scrapper_mcp/       Success: no issues found in 38 source files
pytest tests/unit/               258 passed
pytest tests/integration/        41 passed, 1 unrelated deprecation warning
```

Current GitHub checks are green. `google-re2==1.1.20251105` remains the latest PyPI release. I did not rerun `uv sync --reinstall`, build the Docker image, or repeat the live concurrent `/health` probe in this round.

### What this revision gets right

- The exact `\Q...\E` bypass from round 2 is rejected before RE2 compilation.
- Valid POSIX classes and escaped literal braces retain their documented behavior.
- The four/five variable-range boundary is now explicit and client-visible.
- The scanner is forward-only for malformed braces, removing the prior quadratic suffix rescans.
- Matching-time ReDoS remains closed: client patterns use RE2 with no `re` fallback, a fixed 2 MiB `max_mem`, case-insensitive matching, and suppressed absl stderr output.
- Polish validation/precondition errors, configuration clamping, `ResultStore` injection, whole-set record caps, and the `DocumentStore.re.escape` contract remain correctly wired.

### Findings

#### Critical

1. **`[.` / `[=` nested-token handling creates another compile-time scanner bypass** - see the inline comment on `src/law_scrapper_mcp/services/pattern_matching.py`. A 504-character pattern passes the default length cap, hides 62 expensive ranges from the scanner, and exceeded an independent three-second hard timeout in a killable subprocess. This leaves the unauthenticated synchronous event-loop DoS reachable.

#### Important

1. **The original expensive-range regression still runs in the pytest process** - see the inline comment on `tests/unit/test_services/test_pattern_matching.py`. If the guard regresses, native compilation can monopolize the test worker before its elapsed-time assertion runs.

#### Minor

1. **The one-second subprocess timeout has insufficient Windows scheduling margin** - see the inline comment on `tests/unit/test_services/test_pattern_matching.py`. It has already produced a `TimeoutExpired` during this review, while an immediate rerun completed near the boundary.

### Re-verification of prior findings

- Original unquoted `"a{1,900}" * 62` compile-time DoS: resolved for that spelling; a new bracket-token bypass reaches the same sink.
- Round-1 documentation and error-classification gaps: resolved.
- Round-2 `\Q...\E` bypass: resolved.
- Round-2 POSIX compatibility regression: resolved for valid `[[:name:]]` classes.
- Round-2 threshold/adversarial coverage: four/five and lexical cases were added, but subprocess isolation remains incomplete and the timeout is fragile.
- Round-2 quadratic unmatched-brace scan: resolved.
- The repository's intentional `docs/` policy was handled separately and is not repeated as a finding here.

### Acceptance criteria

Verified from current tests and inspection: 2-12 and 14, except that criterion 1's live `/health` concurrency probe was not repeated. Criterion 13's clean reinstall/container portability was not independently rerun in this round. The broader security goal still fails because accepted input can trigger multi-second synchronous compilation.

### Assessment

**Ready to merge?** No

The specific second-round defects were addressed, but `[.` and `[=` provide a new lexical bypass that still permits attacker-controlled synchronous RE2 compilation beyond an event-loop-safe bound. Close that path and cover all expensive adversaries with killable subprocess regressions before merging.

Note: GitHub rejects `REQUEST_CHANGES` on an author-owned PR, so this final round is submitted as `COMMENT` with the same merge-blocking verdict.

src/law_scrapper_mcp/services/pattern_matching.py
numikel: Fixed in 1f47299. The preflight now consumes only exact RE2 POSIX tokens (`[[:name:]]` / `[[:^name:]]`) from a whitelist. Incomplete `[.` / `[=` forms are ordinary class content, so a real `]` closes the class and subsequent ranges are counted. Subprocess regressions cover both bypass spellings.

tests/unit/test_services/test_pattern_matching.py
numikel: Fixed in 1f47299. The isolated-process timeout is now three seconds, providing Windows process-startup scheduling margin while remaining below the demonstrated multi-second unsafe compilation path.

tests/unit/test_services/test_pattern_matching.py
numikel: Fixed in 1f47299. All known expensive variable-range shapes now use the killable _compile_in_subprocess helper: the direct case, quoted-literal bypass, and both bracket-token bypasses. The pytest worker no longer compiles these inputs directly.

src/law_scrapper_mcp/services/pattern_matching.py
numikel: Fixed in 1f47299. The preflight now recognizes only complete RE2 POSIX tokens ([[:name:]] and [[:^name:]]). Unsupported [./[= forms are ordinary class content, so their first closing bracket exposes later range quantifiers. Subprocess-isolated regressions cover both bypass spellings.

tests/unit/test_services/test_pattern_matching.py
numikel: Fixed in 1f47299. All known expensive compilation adversaries, including the plain `"a{1,900}" * 62` case, now run through `_compile_in_subprocess` with a hard timeout.

tests/unit/test_services/test_pattern_matching.py
numikel: Fixed in 1f47299. Raised `COMPILE_TIMEOUT_SECONDS` to 3 so Windows process startup stays below the threshold while remaining well under the demonstrated multi-second attack compile times.

PR #13, opened 2026-08-10
feat(mcp): migrate to official stateless MCP SDK 2.0.0 and paginated services

numikel:
## Description

This PR completes the migration to the official stateless MCP SDK 2.0.0 and introduces typed services, paginated responses, and transport security controls.

### Key Changes
- **Stateless MCP SDK 2.0.0 Migration**: Migrated server runtime to standard MCP SDK with stateless transports (STDIO and SSE/HTTP).
- **Service Layer Architecture**: Introduced typed \AppContext\ and extracted modular services for date calculations, relationship analysis, act comparisons, and metadata handling.
- **Pagination & Response Limits**: Implemented strict response bounds and pagination contracts across search results, act contents, and stored queries (\page_info\ / \
eturned_count\ guarantees).
- **Transport Security**: Enforced explicit Host and Origin allowlist validation on HTTP/SSE endpoints independent of bind address (\


numikel:
## Code review — outcome

Reviewed `6872e10..dc20618` against the design (`docs/superpowers/specs/2026-08-09-mcp-sdk-migration-sequencing-design.md`) and the plan (`docs/superpowers/plans/2026-08-09-mcp-sdk-migration-sequencing.md`), across four scopes: SDK migration and transports, the pagination contract, the service/adapter layer, and release quality.

**Verdict: the migration is sound.** All eleven completion criteria from the design are met — 13 tools with concrete output schemas, native structured content, `isError=true` on failure, stateless `/mcp`, negotiated protocol `2026-07-28`, and a gate-6 Host/Origin allowlist confirmed by live probe against a server running with the production bind address.

Four Major findings surfaced and are fixed in this branch. Three deserve highlighting because they were invisible to the existing tests.

### Fixed here

| Commit | Finding |
|---|---|
| 1c3f09a | `transport` was a bare `str` and `main()` wrapped it in `typing.cast`, which validates nothing at runtime. `sse` was **valid in 2.4.0**, so an operator carrying `LAW_MCP_TRANSPORT=sse` into 3.0.0 got a server silently started on stdio — HTTP port never opened, healthcheck failed after its start period, nothing in the log. The plan listed this under Task 11; the version bump landed, the validation did not. |
| 3ac44a2 | `SejmApiError` (429/500/400) and `httpx.TimeoutException` were absent from the error classification map, so ordinary upstream conditions were reported to the client as internal server faults and wrote a traceback each time. Fixing it required care: the `SejmApiError` message embeds `e.response.text`, so the new category carries its own fixed message rather than falling through to `str(exc)`. |
| 1d25b42 | **Gate 6 was not guarded on the code path it was written for.** Every transport test built its own ASGI app and supplied `transport_security` itself, and the subprocess fixture ran on `127.0.0.1` — the one address for which the SDK enables validation automatically. Deleting `transport_security` from `main()` failed nothing. |
| 91243e3 | `filter_results` and `track_legal_changes` normalised `limit`/`offset` with a bare `int()`, so a non-numeric value returned CPython's English `invalid literal for int() with base 10: 'abc'` to the client, against the all-Polish message contract. Narrower than it first appeared: negative values were already rejected and over-maximum limits already clamped in the service layer — only the message language was wrong. |

Plus the architectural debt this review was asked to close:

- **9d9b4ff** — `ContentService` extracted from `read_act_content` and `search_in_act`. Both adapters are down to one `await` and no domain computation. Hint behaviour is reproduced exactly, and the integration suite stayed at 70 passing tests, which is the evidence the refactor moved nothing.
- **85e1d22** — the thin-adapter guard rewritten from substring matching to `ast` analysis. Run against the pre-refactor sources it flags both adapters (`act_search::search_in_act` for user-facing f-strings, `act_content::read_act_content` for two awaits and two f-strings); the old substring rules passed both blind.
- **132ca2d**, **5a2be92** — `ruff format --check` added to CI, and the README's Host/Origin parity claim narrowed to `/mcp`.

### Evidence for the two guards

Both new guards were verified by breaking the thing they protect, not just by passing:

```
# 1d25b42 — transport_security temporarily removed from main()
>       assert forged_host.status_code == 421
E       assert 200 == 421
FAILED tests/integration/test_http_transport.py::test_live_server_enforces_allowlist_and_statelessness
# restored -> 1 passed

# 85e1d22 — new rules run against dc20618
dc20618 act_search.py::search_in_act     -> WOULD FAIL: user-facing f-strings at [96]
dc20618 act_content.py::read_act_content -> WOULD FAIL: awaits=2 (max 1); f-strings at [105, 122]
dc20618 act_search.py  -> old guard PASSES (blind)
dc20618 act_content.py -> old guard PASSES (blind)
```

### Quality gates on the pushed head

```
ruff check          -> All checks passed!
ruff format --check -> 79 files already formatted
mypy                -> Success: no issues found in 45 source files
pytest tests/unit/  -> 380 passed, 3 warnings
pytest integration  -> 70 passed
```

From 360 unit / 67 integration / 6 warnings at the start of the review: +23 tests, −3 warnings.

### Worth recording

`LOOPBACK_TRANSPORT_SECURITY` in `server.py` is character-for-character the block the SDK builds for a loopback host (`mcp/server/lowlevel/server.py:740-744`) — the plan's claim of a parity fix with no invented allowlist is accurate.

One deliberate asymmetry is now documented in the README: the SDK applies this validation inside the Streamable HTTP session manager rather than as whole-app middleware, so `/health` is outside it. That is what keeps container healthchecks working when they connect by container name or bridge IP, and it is also why `/health` discloses the server name and version on the published port. FastMCP guarded it; the SDK does not.

Follow-ups in the next comment.

numikel:
## Follow-ups

None of these block the merge. Split by when they need attention.

### Decide before releasing 3.0.0

- [ ] **`click 8.3.0` is vulnerable** — `PYSEC-2026-2132`, fixed in `8.3.3`. It arrives transitively through `mcp[cli]`. The `override-dependencies` block in `pyproject.toml` already pins `cryptography`, `urllib3` and `idna` above their fix thresholds but does not list `click`. One line plus a lockfile refresh. Left undone deliberately — changing `uv.lock` immediately before a release is the maintainer's call, not a review outcome.
  <sub>Verified with `uv run --with pip-audit -- python -m pip_audit` against this venv. A global `pip-audit` run reports false positives here from packages that are not dependencies of this project.</sub>
- [ ] `uv lock --check` is clean (`Resolved 71 packages`), so the lockfile itself is consistent.

### Do immediately after merging

- [ ] **`CLAUDE.md` will document the wrong version.** `scripts/check_release.py` does not exist on this branch — it was added on `main` after the branch point. A trial merge (`git merge-tree`, no write) confirms the scripts and the "Release management" section survive, but the merged `CLAUDE.md` will read `check_release.py --version 2.4.0`. It cannot sensibly be fixed here: that section does not exist on this branch, so adding it would create a conflict where the merge is currently clean.
- [ ] Run `check_release.py --version 3.0.0` for the first time once it exists in the tree.

Otherwise the version sweep is clean — the only remaining `2.4.0` strings are the `CHANGELOG.md` history entry, the historical `README.md` section heading, and working notes.

### Debt worth a separate issue

Ordered by how much each one costs if left alone.

- [ ] **`npx @modelcontextprotocol/conformance@latest` is a floating tag in CI.** The conformance step currently warns and skips, correctly, because the published `0.1.16` bundle accepts no spec version newer than `2025-11-25` while the SDK reports `2026-07-28`. The day upstream adds support, this step flips from warn-and-skip to a mandatory full run against `conformance-baseline.yml`, and every scenario not in that baseline becomes a hard gate — including `resources-list` and `prompts-list`, which a tools-only server may not pass. Pin the version, or note in the baseline which scenarios will need triage on the first real run.
- [ ] **Tool description contract is untested.** `test_tool_descriptions.py` does not assert `meta["tags"]`, the ≥5 examples, or the "Kiedy użyć"/"Kiedy NIE używać" tree. The migration turned FastMCP's first-class `tags={...}` into the SDK's pass-through `meta={"tags": [...]}`; the tags do reach `tools/list` today, but nothing defends them against a future SDK bump. One parametrized test over the 13 names closes this and surfaces the next item.
- [ ] **Five tools have no decision tree**: `search_in_act`, `track_legal_changes`, `calculate_legal_date`, `get_system_metadata`, `analyze_act_relationships`. Not a regression — they lacked it at `6872e10` too — but `CLAUDE.md` states the requirement.
- [ ] **`search_legal_acts` swallows invalid numeric arguments.** It uses `contextlib.suppress`, so `limit="abc"` is ignored and a success is returned. After 91243e3 it is the only tool behaving this way. `search`/`browse` contracts were deliberately excluded from F04–F08, so this is a conscious choice to make, not an oversight to fix blindly: either apply the same validation or document the tolerance in the parameter description.
- [ ] **Default limit `20` is duplicated four times** — `models/pagination.py`, `tools/search.py`, `tools/browse.py`, `services/search_service.py`. Changing one desynchronises the hints from the real page size.
- [ ] **`SERVER_INSTRUCTIONS` promises hints on every response**, but `search_in_act` and `track_legal_changes` return none.
- [ ] **`base_date=""` changed meaning** — it used to mean "today" (`if base_date:` treated the empty string as falsy) and now raises. Intentional and tested, but it is a public argument outside the agreed list of broken contracts. Worth a `CHANGELOG.md` line.
- [ ] **`Host: 127.0.0.1` with no port is rejected (421)** — `transport_security.py` only matches the `"<base>:*"` pattern against hosts containing a colon, where FastMCP normalised first. Irrelevant while HTTP is loopback-only; it will surprise whoever first puts this behind a proxy on port 80.
- [ ] **`test_server_load.py`** at the repo root is neither collected (`testpaths = ["tests"]`) nor linted, and its module list omits everything gates 1–3 added. Delete it or fold it into the suite.
- [ ] **The STDIO test depends on the private SDK symbol `mcp.client.stdio._parse_line`.** It is the cleanest capture point available, but a patch-level SDK bump can break it — worth a comment naming the version it was verified against.
- [ ] **Client-visible error text mixes languages.** The SDK wraps tool exceptions as `f"Error executing tool {name}: {e}"`, so the client sees an English prefix with Polish content. Sanitisation is intact; this is SDK-side and worth documenting rather than working around.
- [ ] Two `PytestWarning`s remain — a module-scoped `asyncio` mark applied to a sync test in `tests/unit/test_context.py`.

### Not covered by this review

The conformance suite was never executed (`npx` fails locally on a corrupted npm cache), so its behaviour was established by reading the published `0.1.16` bundle rather than by running it. `CHANGELOG.md` and `tests/TEST_SUITE_SUMMARY.md` were checked for breaking-change coverage and `sse` removal, but not sentence by sentence.

PR #28, opened 2026-08-21
fix(client): repair resilience layer (F10, F11, F14, F19, F20, F21)

numikel:
## Summary

Repair the resilience layer of `SejmApiClient` to actually implement the retry, circuit breaker, and error translation mechanisms that were declared but broken.

**Audit findings addressed:** F10, F11, F14, F19, F20, F21 (from `docs/mcp-audit-2026-08-06.md`)

### Core Architecture
- Split `_request()` into three resilience layers with clear responsibilities:
  - `_send` — raw HTTP transport
  - `_execute_with_resilience` — retry loop, circuit breaker accounting
  - `_request` — exception type translation
- Add pure `failure_policy.py` module with `classify_failure()` — testable offline
- Redesigned circuit breaker contract: `try_acquire() / release_success() / release_failure() / release_probe()`
- Remove `tenacity` dependency (not used elsewhere; replaced with explicit retry loop)

### Key Fixes
- **F10**: Retry now works for HTTP 5xx (was: intercepted before retry policy could see them)
- **F11**: Transport errors (`ConnectError`, `ReadError`, `ProtocolError`) now handled uniformly
- **F19**: All 5xx errors now count as circuit breaker failures (was: 5xx silently excluded)
- **F20**: One failed operation = one breaker failure (was: three retries = three failures)
  - Hard budget on retry time (configurable, default 45s)
- **F21**: No race condition in HALF_OPEN — probes counted on admission, not completion
- **F14**: Full test suite for circuit breaker state transitions and failure classification

### New Configuration
- `LAW_MCP_API_RETRY_BUDGET` (default: 45.0s) — total retry window per operation
- `LAW_MCP_API_MAX_ATTEMPTS` (default: 3) — retry attempts

### Files Changed
- **New**: `client/failure_policy.py` — pure classification logic
- **New**: `tests/unit/test_client/test_failure_policy.py`
- **New**: `tests/unit/test_client/test_circuit_breaker.py`
- **New**: `tests/unit/test_client/test_sejm_client_resilience.py`
- **Modified**: `client/circuit_breaker.py`, `client/sejm_client.py`, `client/exceptions.py`, `config.py`, `server.py`, `pyproject.toml`

## Testing
- ✅ 1008 unit tests passing
- ✅ `ruff check` — no issues
- ✅ `ruff format --check` — already formatted
- ✅ `mypy` — no errors
- ✅ All 6 audit findings covered by test suite (red on main, green on this branch)

## Notes for Reviewer
- D8 compliance: all tests written first and verified failing on main
- O1 compliance: retry increase balanced by time budget and `Retry-After` respect
- O3 compliance: `ApiUnavailableError extends SejmApiError` — no existing handlers broken
- D10 compliance: circuit breaker contract is synchronous (no `asyncio.Lock` needed; atomic by construction)

Closes #klaster-4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `search_legal_acts` i `browse_acts` zwracają `page_info`; `browse_acts` przyjmuje `offset` (klaster 2, rozszerzenie zakresu)
 - `MetadataOutput.failed_categories` — jawna lista kategorii metadanych, których nie udało się pobrać (klaster 2, L2)
 - `SearchInActOutput.context_chars_requested` / `context_chars_applied` oraz podpowiedź o przycięciu (klaster 2, L4)
+- `LAW_MCP_API_MAX_ATTEMPTS` (domyślnie 3) i `LAW_MCP_API_RETRY_BUDGET` (domyślnie 45.0 s) — nastawy pętli ponowień klienta API (klaster 4)
 
 ### Changed
+- `SejmApiClient` rozdziela żądanie na trzy warstwy (`_send` / `_execute_with_resilience` / `_request`); translacja wyjątków dzieje się poza pętlą ponowień (klaster 4, F10)
+- Wyłącznik obwodu ma kontrakt `try_acquire` / `release_success` / `release_failure` / `release_probe`; licznik sond rośnie w chwili wpuszczenia, nie po zakończeniu żądania (klaster 4, F21)
+- HTTP 500 i 504 podnoszą `ApiUnavailableError` zamiast ogólnego `SejmApiError` — typ zgadza się teraz z klasyfikacją wyłącznika. `ApiUnavailableError` dziedziczy po `SejmApiError`, więc istniejące bloki `except` łapią tak samo jak dotąd (klaster 4, F19/D7)
 - `search_in_act` buduje kontekst i ustala sekcję wyłącznie dla trafień bieżącej strony; mapowanie sekcji korzysta z wyszukiwania binarnego (klaster 2, L1)
 - `get_system_metadata(category="all")` pobiera pięć kategorii współbieżnie zamiast sekwencyjnie, w granicach semafora klienta (klaster 2, L2)
 - Opis `context_chars` w `inputSchema` narzędzia `search_in_act` podaje granicę 2000 znaków i skutek jej przekroczenia (klaster 2, L4)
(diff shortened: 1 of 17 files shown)


numikel:
## Code review — findings and fixes applied

Holistic review of the full branch diff (`fc87c5b..7856a14`) against the design doc and implementation plan for cluster 4, followed by a verification pass over the fixes. All Critical/Major findings are fixed on the branch; the items below are what changed since the version this PR originally described.

### Critical — fixed

**A confirmed breaker failure was silently dropped when admission was refused mid-sequence** (`client/sejm_client.py`).

When `try_acquire()` refused at attempt >= 2, the latched `breaker_failure_seen` was discarded without being booked. Failure scenario: the breaker is HALF_OPEN, our probe receives HTTP 500, frees its slot for the backoff wait, and on the next attempt finds the slot taken by concurrent traffic. The confirmed server failure vanishes — the circuit stays HALF_OPEN and can be **closed by the probes that happened to succeed**, against an API that is demonstrably still broken. That inverts the purpose of HALF_OPEN, where a single failure must drive the circuit back to OPEN.

This was a defect of *task integration*, not of any single task: Task 2 defined `release_failure()` as one operation doing two things (releasing a slot **and** booking a failure), and Task 3 introduced an exit path that needs only the second half.

Fixed by booking the latched failure before raising `ApiUnavailableError`. Safety of calling `release_failure()` without holding a slot was verified against the code: `try_acquire()` can only return `False` in OPEN (where `_release_slot()` is a no-op) or in a saturated HALF_OPEN (where the decrement and the `_state = OPEN` transition are separated by zero `await`, so no task can observe the intermediate value, and every entry into HALF_OPEN zeroes the counter anyway).

Regression test: `test_failure_survives_admission_refusal_mid_sequence`. **Red-phase verified empirically** — with the fix removed the test fails on `assert <CircuitState.HALF_OPEN> is <CircuitState.OPEN>`.

### Additional fixes

- **`httpx.LocalProtocolError` was retried and counted as a server failure** (`client/failure_policy.py`). It sits under `TransportError` in the httpx hierarchy (verified against httpx 0.28.1), but it signals a malformed request built on *our* side. Retrying rebuilds the identical broken request — three wasted round trips to `api.sejm.gov.pl`, against the project's hard politeness constraint — and booking it as a breaker failure could open the circuit against a perfectly healthy API. Now terminal, in the same branch that already handled `UnsupportedProtocol`.
- **`api_max_attempts` / `api_retry_budget` were unvalidated** (`config.py`). `LAW_MCP_API_MAX_ATTEMPTS=0` produced an empty `range(1, 1)`, so the client raised `ApiUnavailableError` without contacting the API and without recording anything on the breaker — a silent degradation to "no attempts". Now `Field(ge=1)` / `Field(gt=0)`, so a bad value fails loudly at startup. Covered by tests in `tests/unit/test_config.py`.
- **Docstrings and comments translated Polish -> English** across `client/failure_policy.py`, `client/circuit_breaker.py`, `client/sejm_client.py`, `client/exceptions.py` and the three new test files. The repo convention is English docstrings (`client/cache.py`, `services/result_store.py`, `services/pagination.py` and the pre-existing test suite all follow it); Polish is reserved for strings an agent actually reads. `client/circuit_breaker.py` had drifted furthest — a single class docstring mixed English and Polish sections. Exception **messages** correctly remain Polish. The convention is now pinned in `CLAUDE.md` so it does not drift again.
- **Stale docstring on `release_failure()`** still claimed it is called only once attempts run out; the fix above added a second call site.
- **Documentation gaps.** `CHANGELOG.md` had no cluster-4 entry at all, despite a removed runtime dependency, a changed exception type for HTTP 500/504 and two new settings. `README.md` advertised `LAW_MCP_API_MAX_RETRIES` as functional — that setting has no production consumer — while the two live knobs were undocumented.

### Test coverage gaps closed

- **Criterion 5.4c** required an operation under *repeated timeouts* to finish within budget, but both budget tests drove HTTP 503; no test carried a real `httpx.TimeoutException` through the client. Added `test_read_timeout_is_cut_by_the_budget`.
- **Criterion 5.6b** required a sweep over every `httpx` exception type; `CloseError`, `WriteTimeout` and `LocalProtocolError` were missing. Added.

### Corrections to this PR's description

The description above overstates a few things and should be read with these corrections:

1. **"Hard budget on retry time"** — the budget is not hard. Only *planned sleeps* are checked against the deadline; the duration of the requests themselves is not bounded, so with a 30 s read timeout the realistic worst case is ~61 s, not 45 s. `README.md` phrases it accurately ("may **plan** to wait"). Closing this properly means passing the remaining time as a per-request timeout — deliberately left as follow-up work.
2. **"F19: 5xx silently excluded"** — only partly true. On `main`, 502 and 503 *did* call `record_failure()`; the excluded codes were 500, 501 and 504+. The CHANGELOG states this correctly.
3. **"1008 unit tests passing"** — now 1018.
4. **"red on main / all tests written first and verified failing on main"** — literally true, but without evidentiary value for part of the suite. None of the four new test files can even be *imported* against `main`, because `failure_policy` does not exist and `try_acquire` / `release_*` / `_delay` / `max_attempts` are all new names. The redness comes from the rename, not from behaviour. Adapting the names by hand, the 404 test, the generic-4xx test, the open-circuit test and all breaker transition tests would **also pass on the old code** — which is fine, since those are regression guards and the explicit ask of criterion 5.6a, but they do not demonstrate a fix.

### Known follow-ups, deliberately not in this PR

- `config.py` still carries the dead `api_max_retries` field; removing a public env var deserves its own change and a deprecation note.
- The retry budget is not a hard wall-clock bound (see correction 1).
- `backoff()` has no jitter, so concurrent tool calls retry in a synchronised wave.
- `test_lifespan_wires_retry_settings_into_the_client` asserts on `inspect.getsource` text, so reformatting the call breaks it without any behaviour change.
- `SejmApiError` interpolates the entire response body into a message that reaches the agent (pre-existing).

```
ruff check src tests             -> All checks passed!          exit 0
ruff format --check src tests    -> 89 files already formatted  exit 0
mypy src/law_scrapper_mcp/       -> No issues found             exit 0
pytest tests/unit/               -> 1018 passed                 exit 0
pytest tests/integration/ (respx suites) -> 78 passed           exit 0
```

**Verdict: ready to merge.** The architecture solves the root causes rather than the symptoms — moving exception translation outside the retry loop makes F10 structurally unable to recur, and the pure classifier makes the whole policy testable offline. The one integration defect is fixed and pinned by a test with a verified red phase.

PR #33, opened 2026-08-23
feat: lifecycle and observability — graceful shutdown and health reporting

numikel:
## Summary

Process lifecycle and observability improvements addressing audit findings F09, F12, F25, and F32.

### What changed

**Lifecycle & Shutdown (F25, F12)**
- Custom uvicorn bootstrap for HTTP transport to control `timeout_graceful_shutdown` (default 15s, configurable via `LAW_MCP_SHUTDOWN_GRACE`)
- Verified lifespan cleanup survives `CancelledError` (container termination path)
- Updated `docker-compose.yml` with `stop_grace_period: 30s` to ensure process has time to finish graceful shutdown

**Health Reporting (F32)**
- `/health` endpoint now reports upstream circuit breaker state in response body
- Always returns HTTP 200 (liveness probe), degradation reported in payload
- States: `closed` (healthy), `open` (API unavailable), `unknown` (startup/shutdown)

**Content Processing (F09)**
- Document conversion (`html_to_markdown`, `pdf_to_text`, `index_sections`) moved to worker threads via `asyncio.to_thread`
- Size limit enforced before conversion: content exceeding `LAW_MCP_DOC_STORE_MAX_SIZE_BYTES` rejected with `ContentTooLargeError` (Polish message + source PDF URL)
- `/health` stays responsive during large document processing

### Testing

- 1046 unit tests passing
- All quality gates: ruff, mypy, format
- New tests: lifespan cleanup, health states, event-loop freedom during conversion, size limit refusal
- Integration tests verify `/health` with circuit state

### Breaking changes

None. HTTP transport bootstrap is internal detail; STDIO (default) unchanged.

### Configuration

New environment variable:
- `LAW_MCP_SHUTDOWN_GRACE` (default 15s) — graceful shutdown window, must keep `stop_grace_period >= 2 × this value`

Modified behavior:
- `LAW_MCP_DOC_STORE_MAX_SIZE_BYTES` now also enforces pre-conversion limit (previously only bounded storage)


numikel:
## Code review — process lifecycle and state observability

Reviewed the full branch, covering integration between the commits, architecture, regressions,
security, test quality, and merge readiness. The review read the SDK sources this branch mirrors
(`mcp 2.0.0`, `uvicorn 0.37.0`) rather than the diff alone.

**Result: 0 Critical, 2 Important, 9 Minor.** Both Important findings are fixed in this PR and
confirmed by a second, independent pass. Verdict: **ready to merge.**

### What holds up well

- **The uvicorn bootstrap loses nothing the SDK was doing.** `build_http_app()` in `server.py` was
  compared against `MCPServer.run_streamable_http_async`
  (`mcp/server/mcpserver/server.py:1055-1088`): the SDK passes only default values for
  `json_response`, `event_store`, `retry_interval`, and `max_request_body_size`, which this code
  deliberately omits. The Starlette app carries its own lifespan, so `_health_state` is populated
  before the server accepts its first connection, and signal handling stays inside
  `uvicorn.Server.serve()`.
- **`test_http_app_pins_security_critical_kwargs` asserts `transport_security` by object identity.**
  With `LAW_MCP_HOST=0.0.0.0` from `docker-compose.yml`, the SDK's DNS-rebinding auto-fallback does
  not engage, so dropping that kwarg would be a silent security regression. This test closes that
  door.
- **Splitting the PDF branch into `try/except/else` in `act_service.py` is a real bug fix**, not a
  refactor: the previous single `except Exception` would have swallowed a size refusal into the
  "no readable content" placeholder.
- **The offload proof is behavioural** — a concurrent coroutine's progress is counted, rather than
  asserting that `asyncio.to_thread` was called.
- **The offload boundary is respected.** `grep to_thread` gives 3 hits in `act_service.py` and 0 in
  `document_store.py`, which relies on the absence of `await` in its critical sections.
  `ContentProcessor` has no `self.` references at all, so sharing one instance across worker threads
  is safe — checked, not assumed.

### Important findings, both fixed here

**1. Silent truncation was still reachable on the PDF path** — fixed in 4fd4cc5

The size guard measured only the *input* to conversion. That bounds the HTML path, where markdownify
strips markup and the result is smaller than its source, but PDF runs the other way: text streams are
Flate-compressed, so a payload under the limit can extract past it. Such a document passed the guard
and reached `DocumentStore.load`, which truncates at an arbitrary offset
(`markdown[: self._max_size_bytes]`) and reports it with nothing but a `logger.warning`. That is the
silent, undetectable loss this branch chose refusal in order to avoid — and the PDF path is the
fallback for acts with no HTML version, which tend to be the oldest and largest ones.

The fix adds a second guard on the conversion *output*, before section indexing, so one limit governs
the whole document path. The closure is structural rather than incidental: `DocumentStore.load` has
exactly one caller in `src/`, the guard sits ahead of it after all branches converge, and both
comparisons use `>`, so there is no off-by-one gap.

This introduces no false refusals. The refusal region (`len(markdown) > limit`) is identical to the
region `DocumentStore.load` previously truncated, because in production both read the same `Settings`
value. Every document that is now refused was previously being cut; none that previously loaded
intact is now rejected. The behaviour change is exactly "silently cut" → "loudly refused".

Regression test: `test_pdf_expanding_past_the_limit_is_refused_not_truncated`. Verified red before
the fix (`Failed: DID NOT RAISE ContentTooLargeError`).

**2. The composition of the offload with the rebuilt HTTP bootstrap was unproven** — fixed in 73ce5d6

The suite had only the unit-level ticker test, which drives a bare `ActService` — no ASGI app, no
`custom_route`, no uvicorn. That left the one interaction this branch actually created untested:
whether `/health` really stays answerable through the rebuilt stack while a document is converting.

The new test in `tests/integration/test_http_transport.py` issues `GET /health` through
`httpx.ASGITransport` on `build_http_app()` while a worker thread is busy converting. Verified to
fail without the offload: the conversion holds the event loop for its full duration, so it completes
before the probe is even issued, and the assertion message says so rather than reporting a timing
miss.

Measured margins: the probe takes 1.0-3.8 ms against a 1000 ms threshold, and five consecutive runs
of the test land within 1.50-1.51 s, dominated entirely by the deliberate `time.sleep(1.5)`.
Regression detection rests on a deterministic assertion, not on the clock.

### Also corrected

- **d358555 / 57ed2d2** — the changelog did not mention that `get_act_details(load_content=True)` now
  returns `is_error=true` for an oversized act. The first attempt at that entry described the
  previous behaviour incorrectly (`is_loaded=false`); checked against `f8aefe4`, there was no size
  guard at all, so such an act used to be reported as `is_loaded=true` with content cut mid-clause.
  The accurate version is also the stronger one.
- **57ed2d2** — the truncation branch in `DocumentStore.load` is now unreachable from production, so
  it carries a note saying as much; silence in that log is not evidence the limit is off. The guard
  comment in `act_service.py` no longer claims markdown is *always* smaller than its HTML source —
  markdownify escapes `_` and `*` and can expand on text dense in those (measured ratios through
  `ContentProcessor.html_to_markdown`: 0.425 on a realistic act shape, 0.787 on the repo fixture,
  1.333 on synthetic escape-heavy text).
- **7cb894d** — `ruff format --check .` was failing on `scripts/check_release.py`, inherited from
  `f8aefe4` on `main` and untouched by this branch. Formatted in its own commit (pure line wrapping,
  no semantics) so it can be dropped if a strictly scoped diff is preferred. The repo-wide format
  gate passes for the first time.

### Quality gates

```
$ python -m pytest tests/unit/ -q                     1047 passed in 56.00s        exit=0
$ python -m pytest tests/integration/ -q -m integration  98 passed in 29.48s       exit=0
$ python -m ruff check src tests                      All checks passed!           exit=0
$ python -m ruff format --check .                     97 files already formatted   exit=0
$ python -m mypy src/law_scrapper_mcp/                no issues in 46 source files exit=0
```

CI on `57ed2d2`: all 9 checks green.

### Minor findings — recommendations, none blocking

Worth considering before the release:

1. **`int(settings.shutdown_grace)` silently truncates fractional values** (`server.py`). Both
   bootstrap tests use whole numbers, so the truncation is invisible to them.
   `LAW_MCP_SHUTDOWN_GRACE=2.5` yields a 2 s window rather than the configured 2.5 s. Making the
   field an `int` rejects fractions at startup instead and drops the cast — `uvicorn.Config` takes
   `int | None` anyway.
2. **A typo in `LAW_MCP_LOG_LEVEL` now aborts startup on the HTTP branch.** `setup_logging` tolerates
   junk, but the new bootstrap hands the value to `uvicorn.Config`, which does
   `LOG_LEVELS[self.log_level]` (`uvicorn/config.py:389`) → `KeyError`. Previously uvicorn read the
   level from the SDK's own settings, so this failure path did not exist. The side effect is positive
   — uvicorn now honours the project's setting — but a `Literal[...]` on the field would close it.
3. **`uvicorn` is imported directly in `server.py` but not declared in `pyproject.toml`.** It is
   present as a core dependency of `mcp`, so nothing breaks today, but the direct import of a
   transitive dependency is debt. (`starlette` has been in the same position for longer.)

For the backlog:

4. **`asyncio.to_thread` workers are outside the `timeout_graceful_shutdown` window.** The ASGI task
   is cancelled when the window expires, but the thread runs to completion and `asyncio.run` then
   waits on the executor join. With `stop_grace_period: 30s` this ends in `SIGKILL`. The ordering is
   favourable, though — uvicorn shuts the lifespan down before collapsing the loop, so the cleanup in
   the `lifespan` `finally` still runs; the consequence is an unclean exit code, not lost cleanup.
   Worth one sentence next to `stop_grace_period` in `docker-compose.yml`, where the `>= 2x` relation
   is currently justified by the uvicorn window alone.
5. **The limit applies after the whole response body is materialised** — `sejm_client.py` returns
   `response.text` / `response.content`, so peak memory stays unbounded despite the limit, and
   `html.encode("utf-8")` adds a transient copy on top. A byte budget applied during download would
   be the real guard; it belongs with the outbound rate-limiting work, same file, same constraint.
6. **`/health` can report `open` indefinitely with no traffic.** `CircuitBreaker.state` is
   deliberately side-effect free; the OPEN → HALF_OPEN transition happens only in `try_acquire()`.
   After the recovery timeout the endpoint still shows `open`, although the next request would be
   admitted as a probe. `README.md` lists `half_open` among the possible values, but
   `tests/unit/test_health_endpoint.py` covers only `closed`, `open`, and `unknown`.
7. **Stale references to `server.main()` as the owner of the security kwargs**, in
   `tests/integration/test_http_transport.py` comments and `tests/TEST_SUITE_SUMMARY.md`.
   `build_http_app()` owns them now. The `asgi_app` fixture also rebuilds those four kwargs by hand
   rather than calling `build_http_app()`, duplicating the production contract in a test.
8. **Nothing pins the ordering of `_health_state.clear()` inside the `lifespan` `finally` block.**
   Clearing it first is deliberate — a populated handle would let `/health` claim a live breaker if
   `client.close()` raised — but a test forcing that exception and asserting an empty handle would
   cost about five lines.
9. **12 of 14 commit subjects exceed 50 characters**; the 72-character hard cap is respected.
   Informational.

PR #35, opened 2026-08-25
feat: logging hygiene — correlation, UTF-8, privacy, timezone (F13, F40, F49, F50)

numikel:
## Summary

This PR implements **four interconnected logging improvements** across the MCP server to achieve production-grade observability while protecting sensitive legal query text:

- **F40**: Request correlation IDs (`request_id` field) allow tracing a single tool call across concurrent requests
- **F13**: User query text no longer leaks to INFO-level logs in result stores or tool error paths
- **F49**: Polish diacritics pass through stderr and JSON unescaped and unmangled
- **F50**: Log timestamps carry explicit UTC designation (`+00:00`) instead of naive values

All four changes are bundled in one PR because they touch the same `logging_config.py` fileâ€”splitting them would create rebasing friction without the granularity that four separate commits already provide within this PR.

## Structure

Five commits (the fifth records the change in CHANGELOG):

1. `fix(logging): force UTF-8 on stderr` â€” Prerequisite for ensuring diacritics survive
2. `feat(logging): add request correlation id` â€” Unique per-tool-call ID, context-local
3. `fix(logging): keep user query text off INFO` â€” Split query into DEBUG-only log records
4. `fix(logging): stamp log timestamps as explicit UTC` â€” Add `+00:00` suffix to JSON timestamps
5. `docs(changelog): record the log output contract change` â€” Update CHANGELOG for observability

## Verification

- âś… 1056 unit tests pass
- âś… ruff check + mypy validation pass
- âś… Backward-compatible: only log output format changes, no API or tool behavior changes
- âś… F13 security: query text never reaches INFO-level persistent storage on stderr
- âś… F40 observability: correlated requests visible in `request_id` field
- âś… F49 & F50: UTF-8 and timezone contracts explicit

## Ready for code review

This PR awaits your review on GitHub before merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **Log output contract** — Every log record now carries a `request_id` field: `lifespan` for records emitted outside a tool call, an eight-character hexadecimal id for records emitted inside one. JSON logs keep Polish diacritics literal instead of escaping them to `\uXXXX`, and the JSON `timestamp` field now ends in an explicit `+00:00` instead of being a naive UTC value. Log aggregation pipelines parsing this output absorb one field-shape change, not three.
+
+### Fixed
+
+- **Query text no longer reaches INFO through the result store or the tool error path** — `ResultStore.store()` logs the result set id and counts at INFO and moves the query summary to DEBUG; the tool error decorator logs validation failures at ERROR without the caller-supplied message, which moves to DEBUG. In the legal domain a query can be sensitive even when every act it matches is public. Tool responses are unchanged.
+
 ## [3.1.1] - 2026-08-24
 
 See [docs/changelogs/v3.1.1.md](docs/changelogs/v3.1.1.md) for details.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- **Log output contract** — Every log record now carries a `request_id` field: `lifespan` for records emitted outside a tool call, an eight-character hexadecimal id for records emitted inside one. JSON logs keep Polish diacritics literal instead of escaping them to `\uXXXX`, and the JSON `timestamp` field now ends in an explicit `+00:00` instead of being a naive UTC value. Log aggregation pipelines parsing this output absorb one field-shape change, not three.
+- **Log output contract** — Log aggregation pipelines parsing this output absorb one field-shape change rather than three:
+  - Every record carries a `request_id` field: `lifespan` outside a tool call, an eight-character hexadecimal id inside one.
+  - JSON logs keep Polish diacritics literal instead of escaping them to `\uXXXX`.
+  - The JSON `timestamp` field ends in an explicit `+00:00` instead of being a naive UTC value.
 
 ### Fixed
 
-- **Query text no longer reaches INFO through the result store or the tool error path** — `ResultStore.store()` logs the result set id and counts at INFO and moves the query summary to DEBUG; the tool error decorator logs validation failures at ERROR without the caller-supplied message, which moves to DEBUG. In the legal domain a query can be sensitive even when every act it matches is public. Tool responses are unchanged.
+- **Query text no longer reaches INFO through the result store or the tool error path** — `ResultStore.store()` logs the result set id and counts at INFO and moves the query summary to DEBUG. The tool error decorator logs `validation` and `upstream` failures at ERROR without the exception text — caller-supplied for the former, an upstream response body that a 4xx can fill with the rejected request's parameters for the latter — keeping both recoverable at DEBUG; an upstream HTTP status still appears on ERROR. In the legal domain a query can be sensitive even when every act it matches is public. Tool responses are unchanged.
 
 ## [3.1.1] - 2026-08-24
 
--- a/README.md
+++ b/README.md
@@ -185,6 +185,18 @@ All settings are configured via environment variables with the `LAW_MCP_` prefix
 | `LAW_MCP_LOG_LEVEL` | `INFO` | Log level: DEBUG, INFO, WARNING, ERROR |
 | `LAW_MCP_LOG_FORMAT` | `text` | Log format: `text` or `json` |
 
+Logs go to stderr (stdout carries the MCP protocol on the STDIO transport). Every record carries a `request_id` correlating it with a single tool call — `lifespan` for records emitted outside one:
+
+```text
+2026-08-24 12:00:00,123 - law_scrapper_mcp.tools.search - [a1b2c3d4] - INFO - Stored result set rs_1: 20 results (total 137)
+```
+
+```json
+{"timestamp": "2026-08-24T10:00:00.123456+00:00", "level": "INFO", "logger": "law_scrapper_mcp.tools.search", "request_id": "a1b2c3d4", "message": "Stored result set rs_1: 20 results (total 137)"}
+```
+
+Set `LAW_MCP_LOG_LEVEL=DEBUG` to recover the detail deliberately kept off INFO: search query text and the exception messages behind `validation` and `upstream` failures.
+
 Example environment configuration:
 
 ```bash
--- a/src/law_scrapper_mcp/logging_config.py
+++ b/src/law_scrapper_mcp/logging_config.py
@@ -2,12 +2,20 @@
 
 from __future__ import annotations
 
+import contextlib
 import contextvars
 import logging
 import sys
 from typing import Literal
 
-request_id_var: contextvars.ContextVar[str] = contextvars.ContextVar("request_id", default="lifespan")
+DEFAULT_REQUEST_ID = "lifespan"
+"""Correlation id carried by records emitted outside a tool call.
+
+Shared by the ContextVar default and both formatter fallbacks so the two output
+formats cannot drift apart.
+"""
+
+request_id_var: contextvars.ContextVar[str] = contextvars.ContextVar("request_id", default=DEFAULT_REQUEST_ID)
 """Correlation id of the tool call in flight.
 
 The default is a value, not a sentinel: records emitted from `lifespan`,
@@ -20,8 +28,9 @@ class RequestIdFilter(logging.Filter):
     """Stamp every record with the current correlation id.
 
     Attached to the handler rather than to the `law_scrapper_mcp` logger on
-    purpose: records from uvicorn and httpx know nothing about this mechanism,
-    and the text format references `%(request_id)s` on every record it renders.
+    purpose: uvicorn and httpx know nothing about this mechanism, yet a record
+    they emit during a tool call belongs to that call. Attaching here is what
+    gives such records the live id instead of the formatter fallback.
     """
 
     def filter(self, record: logging.LogRecord) -> bool:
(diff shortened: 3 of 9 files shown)


numikel:
## Code review — `02061e2..eab45e5`

Reviewed commit-by-commit and in aggregate against `docs/superpowers/specs/2026-08-24-klaster-6.md` and `CLAUDE.md`. All verification commands were run from the branch worktree, not taken on trust from the PR description.

**Verdict: merge with fixes.** No Critical issues.

---

### Verification output

```
uv run pytest tests/unit/ -q              ->  1056 passed in 119.35s
uv run pytest tests/integration/ -q -m integration
                                          ->  98 passed in 101.90s
uv run ruff check src tests               ->  All checks passed!
uv run ruff format --check src tests      ->  95 files already formatted
uv run mypy src/law_scrapper_mcp/         ->  Success: no issues found in 46 source files
```

Baseline collected at `02061e2`: **1047 tests**. The branch runs 1056, i.e. exactly the 9 tests this PR adds. The `190` figure in spec criterion 7 is stale; the number in the PR description is the correct one.

---

### Acceptance criteria (spec section 6)

| # | Criterion | Verdict | Evidence |
|---|---|---|---|
| 1 | F40 / field present | PASS | `logging_config.py:80` always emits the key; the outside-a-call test asserts `"lifespan"`, the concurrency test asserts `re.fullmatch(r"[0-9a-f]{8}")`. |
| 2 | F40 / isolation | PASS | `test_logging_config.py:97-121` — real `asyncio.gather` with `await asyncio.sleep(0)` to force interleaving; asserts two distinct ids and that the parent context stays `"lifespan"`. |
| 3 | F49 / encoding | PASS | Both `text` and `json` covered, asserting `"\\u017c" not in rendered`; the reconfigure test asserts the exact `{encoding: utf-8, errors: backslashreplace}` call. |
| 4 | F13 / result store | PASS | `result_store.py:83` logs id + counts, `:86` moves `query_summary` to DEBUG; test asserts absence across every INFO record. |
| 5 | F13 / error handling | PASS | `error_handling.py:94-95`; `test_error_handling.py:86-89` asserts `all(detail not in r.getMessage())` across ERROR records. |
| 6 | D9 / timezone | PASS | `logging_config.py:77` uses `datetime.now(UTC)`; `rg -n "utcnow" src/` returns no matches. |
| 7 | No regressions | PASS | See above. |
| 8 | Quality gates | PASS | All three clean. |
| 9 | Visibility | PASS | One `Changed` bullet covering all three format changes as an aggregation-contract change, in English, under `[Unreleased]` — correct for D10. |

---

### Strengths

- **The binding commit order holds.** `d783fd9` introduces `_force_utf8_stderr()` and nothing else; only then does `a5959a7` add `ensure_ascii=False` (`logging_config.py:85`). Each commit is independently revertable — `2611704` touches only the two `timestamp` lines plus its own test, so reverting F50 would not take F40 with it. That is exactly the argument in A12.
- **The filter is on the handler, not the app logger** (`logging_config.py:103`), and the docstring at `logging_config.py:22-25` explains why in terms of uvicorn/httpx records rather than restating the code. O4, correctly implemented.
- **`test_polish_diacritics_are_not_escaped` (`tests/unit/test_logging_config.py:124-146`) defends against a real false positive.** A `%(request_id)s`/filter mismatch raises `KeyError` inside `Formatter.format()`, `Handler.handleError()` swallows it and echoes the record into the same stream — so a naive `message in rendered` assertion would pass on a completely broken formatter. The test asserts `"--- Logging error ---" not in rendered_text` first.
- **The `restore_logging_state` autouse fixture (`tests/unit/test_logging_config.py:34-53`)** saves and restores root handlers, root level and the `law_scrapper_mcp` level. `setup_logging()` strips every root handler including `caplog`'s; without this the rest of the session would be quietly corrupted.
- **`tests/unit/test_services/test_result_store.py:482-492`** explains why `caplog.at_level` is scoped to the `law_scrapper_mcp` ancestor rather than the module logger, defending against a cross-module test-order dependency introduced by `test_server_bootstrap.py`. Real hazard, correctly diagnosed.
- **The `.set()`-without-`.reset()` question checks out.** `mcp` 2.0.0 dispatches every JSON-RPC request through `_spawn` -> `anyio` `task_group.start_soon` (`mcp/shared/jsonrpc_dispatcher.py:606,671-686`), so each tool call runs in its own task with a copied context. `.set()` cannot leak to siblings or back to the reader loop. No tool calls another decorated tool, and `comparison_service.py:21-22` uses `asyncio.create_task` inside a call, so children correctly inherit the parent id.
- **F13 covers the regex path properly.** `PatternValidationError` subclasses `ValueError` (`services/pattern_matching.py:63`), so a rejected caller regex — whose message can embed the offending pattern fragment — classifies as `validation` and is redacted from ERROR.

---

### Issues

#### Critical

None. Specifically, the O2 hazard is handled in the new tests: every test in `test_logging_config.py` replaces `sys.stderr` via `monkeypatch`, and `RecordingStream.reconfigure` (`tests/unit/test_logging_config.py:29-30`) is an inert stub, so the process's real stderr is never reconfigured there.

#### Important

**1. `test_server_bootstrap.py:86` and `:104` reconfigure pytest's real captured stderr and never restore it**

`server_module.main()` calls `setup_logging()` (`server.py:290`), which now calls `_force_utf8_stderr()`. Under this project's pytest configuration (default `fd` capture, pytest 9.1.1) `sys.stderr` at that point is `_pytest.capture.EncodedFile`, whose MRO is `['EncodedFile', 'TextIOWrapper', ...]` and which therefore does expose `reconfigure`. The guard at `logging_config.py:44-46` does not trigger, and the call goes through against pytest's capture stream with no restore anywhere.

Important rather than Critical because the observable impact today is nil — pytest already captures as UTF-8, and the only change is `errors` moving from `replace` to `backslashreplace`, unreachable for a UTF-8 target. It still matters because this is the O2 class of global mutation now escaping from a test that has nothing to do with logging, and it will silently apply to any future test exercising `main()`.

Fix: stub `setup_logging` for the two `main()` tests, or `monkeypatch.setattr(sys, "stderr", io.StringIO())` before the call.

**2. `error_handling.py:91-93` states a justification that is false for the `upstream` category, hiding a residual F13 channel**

The comment reads *"Other categories draw on constants defined in this project, so they keep their detail on ERROR."* That is contradicted 45 lines above in the same file — `error_handling.py:46-47`: *"`SejmApiError` embeds the upstream response body, so this category must never fall through to `str(exc)`."*

That guard was applied to the **public message** (`_public_message` returns `_UPSTREAM_MESSAGE`), but the **ERROR log** at `error_handling.py:97-103` still interpolates `exc` for `upstream`. Concretely, `sejm_client.py:197-201` constructs `SejmApiError(f"HTTP {status}: {exc.response.text}")` — the raw upstream body. Any upstream 4xx that reflects submitted parameters therefore reaches persistent stderr at ERROR, which is the exact failure mode F13 exists to close.

Whether to widen the code fix is a scope call (D4 named two paths, not three), but the comment must not assert a safety property the code does not have. At minimum correct it and cross-reference `error_handling.py:46-47`; better, log `category` and `exc.status_code` for `upstream` and drop `str(exc)` to DEBUG, matching the `validation` treatment.

**3. Deviation from D8: "three tests" became seven**

Spec section 2 specifies `tests/unit/test_logging_config.py` as "nowy plik, trzy testy", D8 says "Trzy testy celowane". The file ships seven. This reads as a beneficial deviation — A10's rejection was about broad `setup_logging` coverage (log levels, re-initialisation), still correctly absent — and the extras each cover a spec-stated constraint. Flagged so it can be confirmed as intentional rather than scope drift.

#### Minor

**4. `logging_config.py:80` duplicates the `"lifespan"` literal.** The ContextVar default lives at `logging_config.py:10`; the JSON formatter hard-codes the same string as its `getattr` fallback. If the default ever changes, JSON output silently diverges from text output.

**5. The text formatter has no fallback for a missing `request_id` (`logging_config.py:90`), asymmetric with the JSON path.** The handler filter guarantees the attribute today, so this is robustness only — but Python 3.10+ gives the fix for free: `logging.Formatter(fmt, defaults={"request_id": "lifespan"})`. That removes the whole class of swallowed-`KeyError` failure the diacritics test currently has to guard against.

**6. `logging_config.py:54` — the trailing bare `return` inside the `except` block is redundant.**

**7. `CHANGELOG.md` bullets are single ~70-word paragraphs.** Content satisfies criterion 9 well, but Keep a Changelog readers scan; a lead sentence plus sub-bullets would read better in the eventual v3.1.2 notes.

**8. `README.md:186` documents `LAW_MCP_LOG_FORMAT` but nothing documents the log line shape,** which this PR changes. A two-line sample (one `text`, one `json`) would help the operator who has to update a log parser.

---

### Spec corrections (not implementation gaps)

- **The `server.py` scope question is a spec-internal inconsistency.** Section 2 lists `server.py` as in-scope for setting the lifespan context value, but O3 states plainly that `"lifespan"` comes from the `ContextVar` default and that an explicit `.set("startup")` in `main()` is optional and out of scope — and A5 reinforces it. O3 and A5 are the refined decision; section 2 was not updated to match. The implementation follows O3, and the outcome is correct: lifespan records at `server.py:152` and `:218` render `request_id=lifespan` from the default with no code. D3 likewise mentions a `"startup"` value the implementation never produces. Recommend correcting the spec rather than touching `server.py`.
- **Correct the `baseline: 190` figure in criterion 7 to 1047** so it is not cited as a target later.

### Process note

`.worktrees/` is not in `.gitignore`, so worktree artifacts (`.mypy_cache`, `.pytest_cache`) surface as untracked files in the primary checkout's `git status`.

---

### Assessment

**Ready to merge: with fixes.**

All nine acceptance criteria pass against real observed output, the four findings are implemented as specified with correct and genuinely revertable commit boundaries, and the contextvar/task-isolation design holds up against how the MCP SDK actually dispatches requests. The two Important items are small and localized and neither threatens the design.

numikel:
## Review feedback addressed — `eab45e5..255542b`

Four commits on top of the reviewed range. Verified after the last one: **1058 unit tests passed**, **98 integration tests passed**, `ruff check` / `ruff format --check` / `mypy` all clean.

| Finding | Resolution | Commit |
|---|---|---|
| Important 1 — `test_server_bootstrap.py` reconfigured pytest's captured stderr and stripped root handlers, with no restore | Autouse `isolate_logging_setup` fixture stubs `setup_logging` for the whole file, so any future test exercising `main()` is covered too. Also dropped the now-obsolete reference to this file from the result store test's comment | `6c5cd2b` |
| Important 2 — the `upstream` category still interpolated `str(exc)` into the ERROR log, contradicting `error_handling.py:46-47` | Fixed rather than merely re-commented. `validation` and `upstream` now share `_REDACTED_DETAIL_CATEGORIES`: the exception text drops to DEBUG, and the HTTP status stays on ERROR via `_status_suffix()` as the one part that cannot carry caller input. Two tests: the response-body assertion was extended to cover the log records, and a new case pins the no-status path (`httpx.TimeoutException`) | `8da401b` |
| Important 3 — "three tests" became seven | Intentional, and already declared: the plan's *"Odchylenie od D8, zgłoszone jawnie"* section argues that D8's "three" is a minimum thematic scope, not a cap on test functions, since section 6 demands six independent checks. The spec has been reworded to say so. Now eight, with the one added below | — |
| Minor 4 — duplicated `"lifespan"` literal | Extracted as `DEFAULT_REQUEST_ID`, used by the ContextVar default and both formatter fallbacks | `f41fc0a` |
| Minor 5 — text formatter had no `request_id` fallback | `logging.Formatter(..., defaults={"request_id": DEFAULT_REQUEST_ID})`, plus `test_text_format_survives_a_record_without_the_filter` formatting a record that never passed the filter. The `RequestIdFilter` docstring now states the sharper reason it lives on the handler: third-party records emitted *during* a call should get the live id, not the fallback | `f41fc0a` |
| Minor 6 — redundant bare `return` | Replaced the whole `try/except/return` with `contextlib.suppress(Exception)`, which is also the form SIM105 prefers | `f41fc0a` |
| Minor 7 — CHANGELOG paragraphs | `Changed` split into a lead sentence and three sub-bullets; `Fixed` updated to describe the widened `upstream` redaction | `8da401b` |
| Minor 8 — no documented log line shape | README now shows a `text` and a `json` sample next to `LAW_MCP_LOG_FORMAT`, and points at `LAW_MCP_LOG_LEVEL=DEBUG` for the detail deliberately kept off INFO | `255542b` |
| Spec — `server.py` in scope vs O3 | Confirmed spec-internal, as suspected. The plan had already resolved it in favour of O3 in its criteria-mapping section, and the implementation followed. Section 2 now records `server.py` as struck from scope, and D3 no longer advertises a `"startup"` value that is never produced | — |
| Spec — stale `baseline: 190` | Corrected to 1047 in both the spec and the plan. The plan's per-step absolute counts are now flagged as deltas rather than targets | — |
| Process note — `.worktrees/` missing from `.gitignore` | **Not actioned: the finding was wrong.** The entry exists at `.gitignore:106` and is present on `main` too; `git status` on the primary checkout is clean apart from the deliberately untracked `docs/` tree | — |

Test count moved 1056 → 1058: one new upstream-timeout case and one new text-formatter-fallback case.

`docs/superpowers/` stays untracked, per the plan's closing note that the spec and plan this PR references are deliberately kept out of git.

PR #37, opened 2026-08-30
feat: HTTP transport security — auth, allowlists, rate limiting (4.0.0)

numikel:
## Summary

Closes F17 (no authentication), F18 (non-configurable Host/Origin allowlist), F26 (no inbound rate limiting) from the 2026-08-06 MCP audit.

- Two authentication modes for the Streamable HTTP transport: `bearer` (static shared secret, constant-time comparison) and `oauth` (OAuth 2.1 / OIDC resource server — JWKS + discovery, algorithm-confusion resistant, works with Entra ID, Google, AWS Cognito, Okta, Auth0)
- Configurable `Host`/`Origin` allowlist (`LAW_MCP_ALLOWED_HOSTS`/`LAW_MCP_ALLOWED_ORIGINS`), with startup validation that refuses to widen it unless an auth mode is configured
- Per-client HTTP rate limiting (token bucket, trusted-proxy-aware `X-Forwarded-For` handling, `/health` exempt)
- **BREAKING:** default HTTP bind moved from `0.0.0.0` to `127.0.0.1`; binding beyond loopback now requires an auth mode or the server refuses to start
- Deployment docs, Dockerfile/docker-compose.yml updates, audit report closure

Full detail: `CHANGELOG.md` / `docs/changelogs/v4.0.0.md`.

## Process

Built via subagent-driven development: 8 tasks, each through implementer → parallel spec+quality review → fix rounds where needed, plus a final whole-branch review (caught one CI-breaking test reading a deliberately-untracked file, and 3 documentation/comment accuracy gaps — all fixed in the last commit of this PR).

## Test plan

- [x] `pytest tests/unit/ -q` — 1127 passed
- [x] `pytest tests/integration/ -q -m integration` — 106 passed
- [x] `ruff check` / `ruff format --check` / `mypy` — all clean
- [x] `scripts/check_release.py --version 4.0.0` — passed
- [x] Live `docker compose config` verified (via Podman) — fails loudly without `LAW_MCP_AUTH_TOKEN`, resolves cleanly with it

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ See [docs/changelogs/v4.0.0.md](docs/changelogs/v4.0.0.md) for details.
   or `oauth`. A container configured with `0.0.0.0` and no token refuses to
   start instead of exposing an unauthenticated MCP endpoint.
 
+### Added
+
 - **Bearer and OAuth authentication** — Streamable HTTP now supports static bearer tokens and OAuth resource-server token verification.
 - **Per-client HTTP rate limiting** — Bounds request throughput per client on the HTTP transport.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Changed
+## [4.0.0] - 2026-08-30
 
-- **BREAKING CHANGE:** the default HTTP bind moved from `0.0.0.0` to `127.0.0.1`.
-  A deployment relying on the implicit wildcard bind becomes unreachable until it
-  sets `LAW_MCP_HOST` explicitly — which now also requires an authentication mode.
+See [docs/changelogs/v4.0.0.md](docs/changelogs/v4.0.0.md) for details.
+
+### BREAKING CHANGES
+
+- The default HTTP bind moved from `0.0.0.0` to `127.0.0.1`. A deployment that
+  relied on the implicit wildcard bind becomes unreachable until `LAW_MCP_HOST`
+  is set explicitly.
+- Binding beyond the loopback now requires `LAW_MCP_AUTH_MODE` set to `bearer`
+  or `oauth`. A container configured with `0.0.0.0` and no token refuses to
+  start instead of exposing an unauthenticated MCP endpoint.
+
+- **Bearer and OAuth authentication** — Streamable HTTP now supports static bearer tokens and OAuth resource-server token verification.
+- **Per-client HTTP rate limiting** — Bounds request throughput per client on the HTTP transport.
 
 ## [3.1.2] - 2026-08-25
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **BREAKING CHANGE:** the default HTTP bind moved from `0.0.0.0` to `127.0.0.1`.
+  A deployment relying on the implicit wildcard bind becomes unreachable until it
+  sets `LAW_MCP_HOST` explicitly — which now also requires an authentication mode.
+
 ## [3.1.2] - 2026-08-25
 
 See [docs/changelogs/v3.1.2.md](docs/changelogs/v3.1.2.md) for details.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -4,7 +4,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 ## Project overview
 
-Law Scrapper MCP v3.1.2 is a modular Python MCP server that exposes 13 tools for searching and analyzing Polish legal acts from the Sejm API (`api.sejm.gov.pl/eli/`). Built with the official Python MCP SDK (`mcp[cli]==2.0.0`, `MCPServer[AppContext]`), it supports STDIO (default) and stateless Streamable HTTP at `/mcp` on port 7683.
+Law Scrapper MCP v4.0.0 is a modular Python MCP server that exposes 13 tools for searching and analyzing Polish legal acts from the Sejm API (`api.sejm.gov.pl/eli/`). Built with the official Python MCP SDK (`mcp[cli]==2.0.0`, `MCPServer[AppContext]`), it supports STDIO (default) and stateless Streamable HTTP at `/mcp` on port 7683.
 
 ## Development commands
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,7 +26,10 @@ COPY --from=builder /app/src /app/src
 # Set environment
 ENV PATH="/app/.venv/bin:$PATH"
 ENV LAW_MCP_TRANSPORT=streamable-http
-ENV LAW_MCP_HOST=0.0.0.0
+# No network-host override here (D19): the image inherits the loopback bind
+# from the code, and exposing it requires an explicit `-e` override of that
+# setting together with an authentication mode. One place to configure
+# exposure, not two.
 ENV LAW_MCP_PORT=7683
 
 # Switch to non-root user
--- a/README.md
+++ b/README.md
@@ -159,6 +159,12 @@ docker compose -f docker-compose.yml up
 
 All settings are configured via environment variables with the `LAW_MCP_` prefix:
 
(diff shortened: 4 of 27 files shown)


numikel:
## Whole-branch review against the spec and plan

Reviewed `f457273..980e7c4` — the full contents of this PR — against `docs/superpowers/specs/2026-08-29-klaster-7.md` (19 decisions, 20 acceptance criteria) and `docs/superpowers/plans/2026-08-29-klaster-7.md` (tasks 1–8). The T9/cleanup commits belong to a separate PR and were excluded.

Three parallel reviewers with disjoint lenses — threat model, correctness and regressions, spec conformance — followed by two rounds verifying the fixes. Every finding was re-verified in the main context before being acted on; that step mattered, see below.

**Outcome: no Critical, no authentication bypass. Seven Major findings, all fixed. Acceptance criteria 20/20 met (18 met / 2 partial before).**

> [!IMPORTANT]
> The fixes below are **not in this PR's diff yet** — they sit in nine local commits on `release/4.0.0` above `980e7c4`, awaiting a decision on pushing. Everything stated about them was measured against that local tip.

### Findings

Five were in the PR itself:

| | Finding | Consequence |
|---|---|---|
| **M1** | All five `list[str]` settings were decoded from the environment as JSON by pydantic-settings — inside the settings source, before any model validator ran. `LAW_MCP_ALLOWED_HOSTS='mcp.example.com:*'` died with a bare `SettingsError`. | The documented reverse-proxy recipe in `README.md` could not start, and `LAW_MCP_TRUSTED_PROXIES` was unreachable. F18 shipped a working mechanism with no usable documented path to it. Fail-closed, so Major rather than Critical. |
| **M2** | `/health` is exempt from the limiter, but the SDK mounts `AuthenticationMiddleware` at application level, so `verify_token()` ran for it too. | In `oauth` mode an unknown `kid` forces a full JWKS refetch, making an anonymous `GET /health` loop an unmetered amplifier against the operator's identity provider — the denial of service F26 exists to prevent, entering through the path F26 deliberately exempts. |
| **M3** | `uvicorn.Config` was built without `proxy_headers`; uvicorn defaults it to True and rewrites `scope["client"]` from `X-Forwarded-For` for any peer matching `forwarded_allow_ips` (falls back to `127.0.0.1`). | Since the server binds loopback by default, every local client qualified — no reverse proxy needed. `LAW_MCP_TRUSTED_PROXIES` never fired in that path, contrary to criterion 11. The Docker path was never affected: there the peer is the bridge gateway. |
| **M4** | `PyJWKClient(..., cache_keys=True)` enables PyJWT's second cache tier — an `lru_cache` over individual signing keys with no time expiry. | A key the issuer retired stayed acceptable for the lifetime of the process, and revoking a key at the IdP is the only revocation channel a resource server has. Rotation under an unchanged `kid` was worse: tokens signed with the live key were rejected while the dead key kept working. `LAW_MCP_AUTH_JWKS_CACHE_TTL` only ever governed tier 1. |
| **M5** | The `X-Forwarded-For` value became a bucket key without being validated as an address, and only the *first* such header was read. | A proxy that appends rather than edits produces a second header, so the caller's own value won. Both spellings let one client mint unbounded buckets. |

Two more were introduced or left open by the first fix wave and caught by the verification round:

| | Finding | Consequence |
|---|---|---|
| **N1** | The credential strip lived inside `RateLimitMiddleware`, and `build_http_app()` returns the bare app when `rate_limit_enabled` is false. | A security property inherited a setting documented as being about request budgets: `LAW_MCP_RATE_LIMIT_ENABLED=false` re-armed M2's amplifier *and* removed the metering that had been bounding it. |
| **N2** | Deriving `proxy_headers` from `trusted_proxies` looked like a strict improvement over a hard `False` — it preserved access-log accuracy and `X-Forwarded-Proto`. It was not. | `ProxyHeadersMiddleware` writes the header into `scope["client"]` *without checking it is an address* and runs first, so `_client_key` took a forged string as the peer and never reached its own address check. Measured through the real stack: 500 requests from one host with distinct forged headers produced 500 buckets and no throttling at burst 10. Reverted to unconditional `proxy_headers=False`. |

### Acceptance criteria

Criterion 2 was partial — the D6 rule worked, but the literal environment-variable form died with `SettingsError` instead of the Polish diagnostic. Criterion 15 was partial — the audit report's F26 closure claimed `/health` was rate-limited (it is deliberately exempt, D12) and quoted a middleware order that had already been corrected. Both now hold; the other 18 were verified independently, several on the wire rather than through mocks.

### Verification

Each fix has a confirmed RED before it landed, plus counter-tests guarding against over-correction. The last round used mutation testing and found a real coverage gap: a stripper that skipped the strip *only* on the `rate_limit_enabled=True` branch — the one that actually deploys — passed 1166 tests. The test is now parametrised over both branches and the same mutation fails.

Also closed while there: `ws="none"` on the uvicorn config. Both middlewares pass non-`http` scopes straight through, while starlette's `AuthenticationMiddleware` handles `websocket` and calls the verifier — an upgrade would have reached `verify_token` past both the budget and the strip. It was unreachable only because `websockets` is absent from the environment, which is an accident of the dependency tree rather than a decision.

### Quality gates at the local tip

```
ruff check            All checks passed!                            exit=0
ruff format --check   108 files already formatted                   exit=0
mypy                  Success: no issues found in 51 source files   exit=0
pytest tests/unit/    1152 passed                                   exit=0
pytest integration    16 passed  (loopback: http_auth + transport)  exit=0
check_release 4.0.0   Pre-release verification PASSED               exit=0
```

Tests grew by 25 (1127 → 1152). The Sejm-API-backed integration tests were run once and not repeated — `api.sejm.gov.pl` is a public service run by a state institution and this project treats outbound politeness as a hard requirement.

### Not fixed, recorded with recommendations

None blocks a merge. The notable ones: the configuration table omits the mandatory auth variables while listing their subsidiary knobs; `auth_jwks_cache_ttl` has no lower bound, so `0` rejects every token indefinitely; the `/health` path is spelled in four places without a shared constant; criterion 8's assertion counts a substring with `>=` rather than parsing the tool list. With the limiter disabled the JWKS amplifier remains open on `/mcp`, because PyJWT refetches on an unknown `kid` regardless of `lifespan` — closing that permanently needs a cooldown in `JwtTokenVerifier` rather than anything in the HTTP layer, which makes it a candidate for the outbound-throttling cluster.

### What the review process itself showed

Five of the eight defects share one shape: **a security property verified at a different layer boundary than the one production assembles.** Unit tests construct `Settings(...)` in Python, bypassing `EnvSettingsSource`; they drive the ASGI app directly, with uvicorn's middleware out of the stack; and one covered only the non-default branch. Each of those choices is right on its own, and together they left the operator-facing contract untested exactly where it lives.

Two of the defects were in code that *had* a test. `test_jwks_is_fetched_once_for_two_verifications` passed while measuring the wrong cache tier — its stub replaced `fetch_data` wholesale, and the real one populates the set cache, so tier 1 was inert. `test_xff_from_untrusted_peer_is_ignored` passed because uvicorn was not in the stack. A green test is not evidence; a green test that can go red is.

numikel:
Update to the note in my previous comment: the nine review-fix commits are now **in this PR**. It has gone from 16 commits to 25, `+2074/-50`, with `5a49c2f` at the tip — so the callout saying the fixes were not yet in the diff no longer applies.

Gates at that tip:

```
ruff check            All checks passed!                            exit=0
ruff format --check   108 files already formatted                   exit=0
mypy                  Success: no issues found in 51 source files   exit=0
pytest tests/unit/    1152 passed                                   exit=0
pytest integration    16 passed  (loopback: http_auth + transport)  exit=0
check_release 4.0.0   Pre-release verification PASSED               exit=0
```

The nine commits, oldest first:

```
d6851a7 fix(config): accept comma-separated env list values
7b671b9 fix(http): keep exempt paths away from the verifier
f4473f0 fix(server): let trusted_proxies decide on XFF
61a07cb fix(auth): drop the unbounded per-key JWKS cache
033e0ab docs(readme): correct stale allowlist and scope notes
d13088e fix(server): tie uvicorn proxy trust to trusted_proxies
a8f5034 fix(http): read the last forwarded-for header, not the first
85375a8 fix(http): close three gaps found reviewing the fix wave
5a49c2f fix(server): refuse websocket upgrades, cover both strip branches
```

`f4473f0` and `d13088e` are superseded by `85375a8`, which reverts `proxy_headers` to an unconditional `False`. Reading them in isolation will give the wrong impression of the final behaviour: deriving that flag from `trusted_proxies` looked like a strict improvement and turned out to open a bypass, because uvicorn's `ProxyHeadersMiddleware` writes `X-Forwarded-For` into `scope["client"]` without validating it as an address and runs ahead of this project's own check. The commit messages carry that history deliberately rather than hiding it behind a squash.

PR #42, opened 2026-08-31
fix: cleanup wave from Klaster 7 review findings (4.0.1)

numikel:
## Summary

The deferred half of Klaster 7. Twelve Minor findings were raised while reviewing tasks 1–6 of the authentication work, accepted as real but not allowed to block that release; this branch implements them, plus two findings that surfaced later. No behaviour changes for a configuration that was already correct, so this goes out as **4.0.1**.

Branched off #37 and rebased onto `main` after that PR was squash-merged, so the twelve commits here are only this branch's own work. The code depends on the auth and configuration surface #37 introduced.

## What is in it

**Configuration hardening**
- `trusted_proxies` entries are validated as CIDR at startup instead of blowing up on the first HTTP request, with a Polish message naming the variable and the offending entry. Bracketed IPv6 (`[::1]`) is explicitly rejected there — it is valid in `LAW_MCP_ALLOWED_HOSTS` and not in a proxy list, which is the obvious mistake to make by analogy.
- A binary token file raises a clean `ValidationError` rather than an unhandled `UnicodeDecodeError`.
- The security-boundary validator moved to its own module, so the `Settings` field block reads top to bottom without ~65 lines of validation logic in the middle; the import cycle that extraction initially created is closed in the same branch.

**Token verification**
- A `jwks_uri` discovered over plain `http://` is refused.
- Failures talking to the identity provider log at WARNING rather than INFO — those are outages, not the everyday rejection of a bad token, which stays at INFO.
- `PyJWKClientError` no longer reaches a log line carrying its message. PyJWT builds that message from the `kid` read out of the token's **unverified** header (`jwks_client.py:208-216`), so logging it echoed attacker-controlled text into the operator's logs. That clause now logs the exception type only; the other three exception types in the tuple keep their message, since they describe the response rather than the token.

**Tests**
- Boundary cases that were missing: a 31-byte token, `rate_limit_requests=0`, `rate_limit_window=0`.
- A spy asserting `hmac.compare_digest` is actually on the token comparison path, so a regression to `==` fails loudly rather than silently reintroducing a timing side channel.
- `caplog` assertions filter by logger name instead of catching any stray WARNING.
- The `bearer_app` fixture no longer mutates global module state through `importlib.reload`.
- `httpx2` is a declared dev dependency with its client properly closed, rather than an undeclared transitive import.

## Relationship to #37's review

Two of these were found by the review of #37 and routed here rather than fixed there: the `kid` leak (introduced on this branch, absent at #37's base) and the CIDR validation gap. In the other direction, #37's fix making list settings readable from flat environment values is what makes the CIDR validator reachable at all — before it, `LAW_MCP_TRUSTED_PROXIES=10.0.0.0/8` died in the settings source before any validator ran.

The one semantic conflict from rebasing onto #37's tip is resolved in `c55fbda`: `trusted_proxies` is excluded from the generic flat-value parametrize, since it now narrows what counts as a valid entry, and a dedicated test covers the same decoding path with a CIDR-valid value.

## Test plan

- [x] `pytest tests/unit/ -q` — 1164 passed
- [x] `ruff check src tests` — All checks passed
- [x] `ruff format --check src tests` — 110 files already formatted
- [x] `mypy src/law_scrapper_mcp/` — no issues in 53 source files
- [x] `scripts/check_release.py --version 4.0.1` — PASSED

All five re-run after the rebase onto `main`, not carried over from before it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [4.0.1] - 2026-08-31
+
+See [docs/changelogs/v4.0.1.md](docs/changelogs/v4.0.1.md) for details.
+
+Cleanup release — hardening and test coverage from Klaster 7 review findings, no behavior change for existing correct configurations.
+
+### Fixed
+
+- **`trusted_proxies` CIDR validation** — Moved to startup instead of first request; binary `LAW_MCP_AUTH_TOKEN_FILE` files now produce a clean error instead of `UnicodeDecodeError`.
+- **OAuth JWKS discovery** — Rejects non-`https://` discovered JWKS URIs; JWKS/IdP communication failures now log at WARNING instead of INFO (including a fix for an except-clause ordering bug).
+
+### Changed
+
+- **Test coverage** — Added token boundary, rate-limit-zero, and timing-safety tests; fixed `caplog` scoping.
+- **Dependency and fixture hygiene** — `httpx2` is now an explicit dev dependency with its test client properly closed; the `bearer_app` test fixture no longer mutates global module state.
+- **Config module split** — Security-boundary validation extracted from `config.py` into `config_validation.py` and `config_primitives.py`. No behavior change.
+
 ## [4.0.0] - 2026-08-30
 
 See [docs/changelogs/v4.0.0.md](docs/changelogs/v4.0.0.md) for details.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -4,7 +4,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 ## Project overview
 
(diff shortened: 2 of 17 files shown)


numikel:
## Security review (code-review high --fix)

Reviewed the diff on this branch. Two findings, both fixed in a follow-up commit pushed to this PR:

**1. OAuth JWKS downgrade bypass — HIGH**
`src/law_scrapper_mcp/config_validation.py`, `enforce_security_boundary` (oauth branch)

`auth_issuer` / `auth_jwks_uri` are typed `AnyHttpUrl`, which accepts plain `http://`. The existing https-only check in `jwt_verifier.py::_discover_jwks_uri` only guards a *discovered* jwks_uri — it never runs when `auth_jwks_uri` is configured directly (discovery is skipped), and it does nothing about the discovery *request* itself going out over plain HTTP against an insecure `auth_issuer`. Either path lets a network attacker MITM the identity-provider communication.

Fix: enforce `https` on both `auth_issuer` and `auth_jwks_uri` directly in `enforce_security_boundary`, so the boundary check catches both paths at startup instead of relying on an assumption documented only in a comment.

**2. Wildcard-bind regression guard dropped during the config.py split**
`tests/unit/test_auth_settings_validation.py`

`test_config_contains_no_wildcard_bind` scanned config source for a literal `0.0.0.0`. When Task 6 split `config.py` into `config_validation.py` / `config_primitives.py`, the logic moved but this test did not — it was deleted with no equivalent added for the new modules, so the wildcard-bind regression guard silently stopped covering the security-boundary code.

Fix: restored the test, extended to scan all three modules (`config.py`, `config_validation.py`, `config_primitives.py`).

Verified locally: `pytest tests/unit/` (1167 passed), `ruff check`/`format` clean, `mypy` clean on the changed file.

PR #43, opened 2026-09-01
feat: bound egress traffic to api.sejm.gov.pl (Klaster 8)

numikel:
## Summary

Bounds the outbound HTTP traffic this server sends to `api.sejm.gov.pl` — a public, state-run API — in four independent dimensions (findings F27, F30, F52, F55 from the 2026-08-06 audit):

- **Egress rate limiting** — a new token-bucket `RateLimiter` (`client/rate_limiter.py`) fronts `SejmApiClient`, configurable via `LAW_MCP_API_RATE_PER_SECOND` (default 5.0) and `LAW_MCP_API_RATE_BURST` (default 10). A `Retry-After` on a 429 now pauses the whole client rather than each failing request separately, capped at 60s so a large/misconfigured header can't silently wedge every tool call.
- **Concurrency fairness** — the single concurrency semaphore is split into `RequestClass.LIGHT` (8 slots, JSON) and `RequestClass.HEAVY` (2 slots, HTML/PDF downloads), so a run of document fetches can't starve concurrent searches. Peak stays at 10 (`LAW_MCP_API_MAX_CONCURRENT` default lowered 10→8, new `LAW_MCP_API_MAX_CONCURRENT_CONTENT`=2).
- **Volume** — `browse_acts` switches from `acts/{publisher}/{year}` (ignores pagination, always returns the entire year — 1.09MB / 1984 records for DU/2024) to `acts/search` with `limit`/`offset` (properly windowed server-side). Verified against production: one real request, 25KB instead of 1.09MB for the default page.
- **Identity** — a regression test pinning the `User-Agent` header actually sent on the wire (the underlying fix already shipped in v3.1.0; this only closes the missing guard).

No MCP tool changes signature, parameter names, or response shape.

## How this was built

Implemented via subagent-driven development: 6 tasks, each with a fresh implementer, then two independent Opus 5 reviewers per task (spec-compliance and code-quality, run in parallel), then a fix loop where needed. Closed with a whole-branch review (Opus 5) that caught 5 cross-task findings no single task's review could see (settings-wiring guard, pause-logging completeness, D8 cache-key coincidence at offset=0, and two documentation gaps) — all fixed and re-verified in a final pass.

## Test plan

- [x] Full unit suite: 1206 passed (baseline on `main` was 1152)
- [x] Integration suite (`-m integration`): 107 passed
- [x] Transport suites (STDIO + HTTP, `-m integration`): 9 passed
- [x] Manual production probe: `browse_acts` against the real Sejm API confirms exactly 1 request, 25,430 bytes, `returned=20 total=1984`


numikel:

Reviewed `e369264..f53972b` against the Klaster 8 spec and plan through three independent lenses: plan alignment / architecture / cross-task integration, egress security, and test quality. Result: **0 Critical, 12 Important**. All Important findings are fixed in `9c9c382`, `587aa49`, `18475ee`, and the fixes were confirmed by a second review pass.

The strongest signal was convergence: two reviewers independently landed on the same layering defect, and one brought a working proof of concept.

**Pacing was not bounded by the operation's time budget.** `acquire()` was awaited inside `_send`, below the layer owning the deadline, so `api_retry_budget` bounded nothing on the success path — a call entering a 60 s pause set by another caller returned after ~90 s under a 45 s budget. In HALF_OPEN a breaker probe slot was held for the whole pause while everyone else got "API unavailable", when the real condition was self-imposed pacing. Measured offline: budget 0.001 s, seven concurrent *successful* calls, last returning after 3.04 s → **0.001 s after the fix**, with clean `ApiUnavailableError`s instead of late successes.

`acquire()` now takes a deadline and refuses *before* starting a wait that would overshoot it, re-checked before every wait since another task can extend a pause mid-wait. I deliberately did not wrap `_send` in `asyncio.timeout` as first suggested: that also cancels in-flight HTTP requests that are merely slow, changing behaviour in code this branch never touched. No new config knob — `api_retry_budget` already carries the right meaning.

**A pause ended by releasing a full burst.** `_refill` is skipped while paused and `_updated` kept its pre-pause value, so the first refill afterwards treated the whole quiet window as elapsed time and clamped the bucket to a full burst. A drained bucket plus a 30 s pause released ten requests back-to-back the instant the server's quiet window ended — precisely the herd D2 exists to prevent. Tokens earned *before* the pause still survive, because that burst is D1's happy path; only tokens that would have accrued *during* the silence are gone. Verified by mutation: removing the fixed line turns the new regression test red.

**`Retry-After` in date form silently disabled the headline feature.** Only the delta-seconds form was parsed, so an RFC 9110 date returned `None` and the client-wide pause never engaged at all — the failing request still backed off, but every other caller kept full pace at a server asking for quiet. Duplicate headers joined by httpx into `"60, 120"` had the same effect.

**`browse_acts` forwarded an unclamped `limit` upstream.** The one behaviour this branch made *worse*: `acts/{publisher}/{year}` ignored `limit`, so exempting the tool from the 100-item clamp cost nothing; `acts/search` honours it, so `browse(limit=100000)` asked api.sejm.gov.pl for a 100,000-record page in a wider record format. Now clamped through the project's own helper, with `offset` validated the same way.

Plus seven test-quality fixes: a behavioural test for the `Settings` → client wiring (it was proven by `inspect.getsource` string matching, which passes for code that never runs), an end-to-end test that a 429 actually holds back a request that never saw it (all three existing tests mocked `pause_for` away), and replacement of a magic 20-yield synchronisation coupled to httpx/respx internals.

### Two documented decisions were reversed, with the maintainer's approval

Both are recorded in code and CHANGELOG rather than slipped in, and tracked in #48:

- `failure_policy.py` was frozen by the plan as a pure, clock-free module. Honouring the date form needs a clock; it is confined to a single `_now()` function, and the module docstring states the deviation.
- Decision P5 exempted `search_legal_acts` **and** `browse_acts` from the shared clamp. That exemption was free for browse only while the year endpoint ignored `limit`. `browse_acts` leaves the exemption; `search_legal_acts` keeps it, since nothing changed there.

```
ruff format --check src tests               114 files already formatted     exit=0
mypy src/law_scrapper_mcp/                  Success: no issues, 54 files    exit=0
pytest tests/unit/                          1217 passed                     exit=0
pytest tests/integration/ -m integration     109 passed                     exit=0
pytest transport -m integration                9 passed                     exit=0
pip-audit                                   No known vulnerabilities found  exit=0
```

Unit tests 1206 → 1217, integration 107 → 109.

Not merge-blocking, filed so they are not lost: #44 (no per-caller share of the egress bucket), #45 (deadline and limiter coupled by convention), #46 (`MAX_SERVER_PAUSE` not operator-configurable), #47 (rate settings accept `inf`), #48 (spec deviations to record), #49 (test harness hygiene), #50 (browse coverage gaps), #51 (F30 has no live-API confirmation).

### Assessment

**Ready to merge.** The plan was followed closely, all of D1–D8 hold, no commit undid an earlier one, and no outbound path bypasses the limiter or the lane split. Worth noting before merge: acceptance criterion 17 asks for confirmation against the live API, and that probe left no artifact in the branch — the browse fixture was written by hand from the spec rather than captured (#51).

numikel:
## Acceptance criterion 17 closed with evidence (`bdaf86d`)

The one part of this branch that rested on trust rather than on an artifact in the repo is now closed. The live probe was re-run by hand — three requests, spaced, with the project User-Agent — and the page it returned is committed as the browse fixture, with a provenance note giving date, URL, sizes and refresh policy.

| Request | Result |
|---|---|
| `acts/search?publisher=DU&year=2024&limit=5` | 6 452 B, `count=5`, `totalCount=1984`, 5 items |
| `acts/search?...&limit=5&offset=5` | 6 769 B, 5 items, zero overlap with page 1 |
| `acts/DU/2024?limit=5&offset=5` | **1 093 220 B, 1984 items** — both parameters ignored |

F30's three claims are now measured, not assumed: `acts/search` honours `limit`, honours `offset` (pages are contiguous and descending by `pos`, nothing skipped or repeated), and the year endpoint ignores both. **169x fewer bytes** at the recorded page size, matching the figures in the spec. Criterion 10 holds too — search returns 27 fields per act against the year endpoint's 15, and the 15 are a subset, so the switch loses nothing while `_format_act`'s whitelist keeps the extras out of the tool response.

Two things the recording changed that a hand-written fixture could not have surfaced:

- The old fixture carried a **`dateEffect` key that neither endpoint returns**, and omitted 15 keys that upstream does return. `_format_act` reads `item.get("dateEffect")`, so `effective_date` has always been `null` in practice. That predates this branch, but the invented fixture had been quietly asserting the opposite — filed as its own issue.
- The positional assertion in `test_the_api_window_is_not_sliced_a_second_time` now derives from the fixture instead of copying values out of it, so refreshing the recording does not cost a test edit.

Gates after the swap: `ruff check`, `ruff format --check`, `mypy`, 1217 unit, 109 integration — all exit 0.

Closes the last open recommendation from the review. #51

Issue #18, opened 2026-08-19
Unify limit and offset validation in search_legal_acts and browse_acts


numikel:
Found during the review of #15. Cluster 2 brought eleven tools under one pagination contract; these two still validate their page parameters differently from the rest.

## Current behaviour

`src/law_scrapper_mcp/tools/browse.py` parses both parameters through `contextlib.suppress(ValueError, TypeError)`:

- `browse_acts(offset="abc")` silently yields page one instead of an error.
- `browse_acts(offset=-5)` is clamped to zero in `SearchService._build_output` and also yields page one, silently.

Every other listing tool raises a Polish `ValueError` from `parse_non_negative` in `src/law_scrapper_mcp/services/pagination.py` for the same input. PR #15 added tests pinning that behaviour for `list_loaded_documents` and `list_result_sets`, which makes the divergence explicit: thirteen tools, two validation contracts for the same parameter name.

Worse in one specific case: `search_legal_acts(offset=-5)` forwards `-5` **to the Sejm API**, because `SearchService.search` builds params with `if offset:` and `-5` is truthy.

## Why it was left alone in #15

The lenient parsing is pre-existing for `limit`, and the plan's P5 resolution deliberately kept these two tools free of the shared clamp so that calls valid in v3.0.0 keep working. Tightening `limit` would change what an already valid call returns.

## Suggested scope

- Route `offset` through `parse_non_negative` in both tools. `offset` on `browse_acts` is **new in #15**, so there is no backward-compatibility cost there.
- Stop forwarding a negative `offset` upstream regardless of the decision above.
- Decide separately whether `limit` should follow. That one is a breaking change and belongs to a major release, with a migration note.

PR #53, opened 2026-09-02
feat: add result-set scope to prevent false negatives (Klaster 9)

numikel:
## Summary

Introduces **result-set scope** — every stored result set (from `search_legal_acts`, `browse_acts`, `filter_results`, `track_legal_changes`) now declares whether it is `complete` (the whole answer to its query) or a `page` (a window cut from a larger corpus). Goal: stop the calling model from concluding "no such act exists" off an empty `filter_results` match over a 20-record window into a 1984-record corpus.

- **Scope lives in one place** — `SetScope`/`ResultSetScope` types, with the derivation rule centralized in `ResultStore.store()` (`scope = complete ⟺ window_offset == 0 and len(results) == total_count`). `filter_and_store` **inherits** the source set's scope instead of re-deriving it, and sets `no_match_is_inconclusive` when an empty match happened over a `page` set.
- **D1 fix** — `search_legal_acts` used to read `count` (page size) into `total_count`, so a query matching 1984 acts reported `20`. Now reads `totalCount` (corpus size), falling back to `count` for responses that omit it.
- **D7 fix** — `search()` only sent `limit` upstream when the caller supplied one; without it the Sejm API built its own default page (measured: 500 records / 709,437 B for `DU/2024`, of which 480 were discarded locally). `limit` is now always sent, defaulting to 20.
- **F48 fix** — pagination hints hard-coded `tool="search_legal_acts"` regardless of which tool actually produced them, so `browse_acts` callers were told to call a different tool for the next page. `search_hints()` now takes `tool_name`/`next_call_params` as required keyword-only parameters (no default — a default would silently reintroduce the bug).
- Scope is surfaced on `SearchOutput`, `ChangesOutput`, `ResultSetInfo`; `filter_results`'s own hints moved into the same hint-builder layer as `search_hints`, and warn explicitly when an empty match proves nothing.

No MCP tool changes signature, parameter names, or the substantive content of its responses (same records, same fields, same order).

**Known follow-up, deliberately out of this scope:** `track_legal_changes` still doesn't send `limit` or read `totalCount`, so a very wide date range (>500 matching acts) could be silently truncated and still get labeled `complete`. This is the same pattern as a previous finding (F30) applied to a wide date range, and needs its own decision (also touching whether `filter_max_records=100` is the right ceiling for a "complete" set) rather than a fix folded into this PR.

## How this was built

Implemented via subagent-driven development: 7 tasks, each with a fresh implementer (Sonnet 5), then two independent Opus 5 reviewers per task (spec-compliance and code-quality, run in parallel), with fix loops where needed. Closed with a whole-branch review (Opus 5) that independently re-ran all quality gates, confirmed the "scope lives in one place" architecture holds end-to-end (exactly one derivation site, one inheritance site across the whole diff), and caught one stray `pytest-asyncio` warning (fixed) plus several documentation/wording items now tracked as follow-ups rather than blocking.

## Test plan

- [x] Full unit suite: 1252 passed (baseline on `main` was 1152), zero warnings
- [x] Integration suite (`-m integration`): 109 passed, live network
- [x] `outputSchema` pinning test confirms the new fields (`result_set_scope`, `no_match_is_inconclusive`, `scope`) reach MCP clients on all five affected tools without dropping any existing field

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,8 +30,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `search_legal_acts` now sends `limit` upstream even when the caller does not supply one
   (default 20). Without it the API built a page of its own choosing — a measured 709 437 B
-  and 500 records for `DU/2024` — of which nineteen out of twenty records were discarded
-  locally. An explicit `limit` is still not clamped.
+  and 500 records for `DU/2024` — of which twenty-four out of twenty-five records were
+  discarded locally. An explicit `limit` is still not clamped.
 - Large searches now return `was_truncated=true` together with a pagination hint, which is a
   consequence of the `total_count` fix rather than a separate behaviour change.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `search_legal_acts` reported the size of the returned page as `total_count`, so a search
   matching 1984 acts answered `20` with `was_truncated=false` and no pagination hint at all.
   It now reads `totalCount` and falls back to `count` for responses that omit it (F31).
+  Clients that relied on `total_count` carrying the page size will see a different value.
+  This is shipped as a fix in a MINOR release, not as a breaking change in a MAJOR one:
+  the field always declared the size of the match set and no documentation ever promised
+  the page size, so the previous value was a defect rather than a contract.
 - The pagination hint from `browse_acts` pointed at `search_legal_acts`, sending the model to
   a different tool than the one it had called (F48).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,34 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Result-set scope (`result_set_scope`, and `scope` in `list_result_sets`) telling the caller
+  whether a stored set is the complete answer to its query or a window cut from a larger
+  corpus. `filter_results` inherits the reach of the set it filtered rather than deriving a
+  new one, and reports `source_scope` alongside it.
+- `no_match_is_inconclusive` on `filter_results`: set when a filter over a windowed set
+  matched nothing, so an empty result is not read as proof that no such act exists.
+- Pagination hints now carry the full next call — the tool that produced them plus every
+  source criterion, `limit` and `offset` — instead of an unparameterised suggestion.
+
+### Fixed
+
+- `search_legal_acts` reported the size of the returned page as `total_count`, so a search
+  matching 1984 acts answered `20` with `was_truncated=false` and no pagination hint at all.
+  It now reads `totalCount` and falls back to `count` for responses that omit it (F31).
+- The pagination hint from `browse_acts` pointed at `search_legal_acts`, sending the model to
+  a different tool than the one it had called (F48).
+
+### Changed
+
+- `search_legal_acts` now sends `limit` upstream even when the caller does not supply one
+  (default 20). Without it the API built a page of its own choosing — a measured 709 437 B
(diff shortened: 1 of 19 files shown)


numikel:

Reviewed `c2ab0c8..5ae52d2` against the Klaster 9 spec and plan through three lenses: plan alignment / architecture / cross-task integration, egress and security, and test quality. Result: **0 Critical, 1 Important**. The Important finding is fixed in `5ae52d2` and the fix was confirmed by a second review pass. All 20 acceptance criteria of spec §6 were checked one by one: 19 hold with a test or code anchor, criterion 19 (integration suite against the live API) rests on CI rather than a local re-run.

**The MAJOR-vs-MINOR argument was missing from the CHANGELOG.** Spec §9 requires it to be recorded there, not only in the spec: `total_count` in `search_legal_acts` now returns a genuinely different value (corpus size instead of page size), which is an honest case for a MAJOR bump. The `Fixed` entry now states the value change and why it ships as a fix in a MINOR release — the field always declared the size of the match set and, checked against `c2ab0c8`, it had no field description at all before this branch, so no documentation ever promised the page size.

### What held up under scrutiny

- **Scope is decided in exactly one place.** `ResultSetScope` is constructed only in `result_store.py`; `store()` is the single derivation site, `filter_and_store` the single inheritance site. Verified by grep over `src/`, not by reading the PR description.
- **`filter_and_store` reads the source set once**, so the tuple return from `store()` does what the plan promised: no second `get()` refreshing `last_accessed` and corrupting LRU order.
- **The `corpus_count is None` trap (spec §8, risk 2) is guarded in code**, not just in the spec: the `store()` docstring says outright not to "fix" that `None` with the source corpus size.
- **A real D9 gap was caught mid-implementation** (`99b8e21`): `browse_acts`'s own `MAX_ITEM_LIMIT` can undercut `filter_max_records`, so a complete-set hint could name a `limit` the tool would silently clamp back into a page. Now capped at whichever ceiling binds, with two dedicated tests.
- **The complete-set hint is executable at any offset.** `next_call_params` in `tools/search.py` and `tools/browse.py` carry source criteria only — no `limit`, no `offset` — so the deliberate "no offset key" comment in `_complete_set_hint` describes the truth. This was checked independently of the reviewer.
- **Zero removed or weakened assertions** across the diff; `test_defaults_match_the_previous_behaviour` (spec §8, risk 5) is byte-for-byte untouched.
- **Egress went down, not up.** `search()` without `limit` used to receive the API's default page (measured 500 records / 709 437 B for `DU/2024`, provenance file committed) and keep 20; it now asks for 20. `filter_results` still makes zero HTTP calls.

### Deviations from spec and plan

All three are intentional and documented, none is drift: the `max_result_limit` parameter above; the AND-logic hint on empty results now appears only when `keywords` was actually passed, so `browse_acts` no longer receives a suggestion it cannot execute; and the corrected `ResultSetTooLargeError` message names the real field `result_set_scope` rather than the plan draft's `set_scope`, which never existed.

### Quality gates after the fix

```
legacy fastmcp remnants                     none                            exit=0
ruff format --check src tests               115 files already formatted     exit=0
mypy src/law_scrapper_mcp/                  Success: no issues, 54 files    exit=0
pytest tests/unit/                          1252 passed, 0 warnings         exit=0
pip-audit                                   No known vulnerabilities found  exit=0
```

Unit tests 1152 → 1252. The integration and transport suites were deliberately not re-run locally (they hit `api.sejm.gov.pl`); CI runs both on every push to this branch.

Not merge-blocking, filed so they are not lost: #54 (`track_legal_changes` fetches an unbounded date range and labels it `complete` — the follow-up spec D6 left open, together with whether `filter_max_records=100` is the right cap for a complete set), #55 (typed `outputSchema` pinning instead of substring search; explicit `limit=0` now floored to `1` without a dedicated test, related to #18).

### Assessment

**Ready to merge.** D1–D9 hold, plan resolutions R1–R3 were followed to the letter (`tools/changes.py` untouched, no new live request), and no commit undoes an earlier one. The version bump to 4.1.0 is intentionally left to `/prepare-release`. The last two commits were rewritten to drop attribution trailers before this comment; their content is unchanged.

Issue #55, opened 2026-09-02
Test polish after Klaster 9: typed outputSchema pinning and explicit limit=0


numikel:
Minor findings from the Klaster 9 review (PR #53). Not merge-blocking; filed so they are not lost.

- `tests/unit/test_server.py::test_scope_fields_reach_the_output_schemas` pins the new fields (`result_set_scope`, `no_match_is_inconclusive`, `scope`) by substring search in `json.dumps(output_schema)`. That proves presence, not type: a regression that typed `result_set_scope` as `str` instead of the nested object would pass. Walk `output_schema["properties"]` and assert the `$ref`/`type` per tool instead.
- `SearchService.search()` now floors the upstream `limit` with `max(limit if limit is not None else DEFAULT_ITEM_LIMIT, 1)` (`services/search_service.py:135`). Before #53 an explicit `limit=0` was falsy and sent no `limit` at all (API default page); now it sends `limit=1`. The change is reasonable and commented, but it sits outside D7's stated scope (the unset default) and has no dedicated test. Pin `limit=0 -> limit=1` with a `respx` assertion, or fold it into #18, which already covers unifying `limit`/`offset` validation across `search_legal_acts` and `browse_acts`.

Issue #54, opened 2026-09-02
track_legal_changes fetches an unbounded date range and labels it complete


numikel:
Follow-up from the Klaster 9 review (PR #53), spec decision D6.

`track_legal_changes` still sends no `limit` to `acts/search` and does not read `totalCount`. For a date range matching more acts than the API's default page (measured: 500 records for `DU/2024` without `limit`, see `tests/fixtures/search_default_page.provenance.md`), the response is silently truncated upstream, stored whole, and — because `window_offset == 0` and `len(results) == total_count` where `total_count` is `len(results)` — labelled `result_set_scope.scope = "complete"`. That is precisely the false "this is everything" signal Klaster 9 removed from `search_legal_acts` and `browse_acts`.

Two questions to settle together, deliberately left out of #53:

1. Should `track_legal_changes` send `limit`/read `totalCount` like the other two producers (same shape as F30), or keep fetching the whole range so that `filter_results` has one honest complete-set source? Spec D6 chose the latter but flagged wide ranges as the open case.
2. Is `filter_max_records=100` (`LAW_MCP_FILTER_MAX_RECORDS`, floor 1, no ceiling) the right cap for a set that declares itself `complete`? A complete set larger than the cap is unfilterable, which undercuts the guarantee `ResultSetTooLargeError` describes.

Acceptance sketch: a range whose `totalCount` exceeds the fetched page must not be stored as `complete`; either it becomes `page` with a truthful `corpus_count`, or the tool paginates upstream.

Issue #52, opened 2026-09-01
effective_date is always null in search and browse output


numikel:
`ActSummaryOutput.effective_date` is part of the response of `search_legal_acts` and `browse_acts` at `standard` and `full` detail levels, and it is **always `null`**.

`SearchService._format_act` reads it as:

```python
output.effective_date = item.get("dateEffect")
```

A live probe on 2026-09-01 (see `tests/fixtures/browse_page.provenance.md`) enumerated the fields both list endpoints actually return:

- `acts/search` — 27 fields per act, **no `dateEffect`**
- `acts/{publisher}/{year}` — 15 fields per act, **no `dateEffect`**

So the `.get()` default has been silently winning on every call. This predates the Klaster 8 endpoint switch — both the old and the new endpoint lack the key — and it stayed invisible because the test fixture was written by hand and *invented* a `dateEffect` key, so the tests agreed with the model rather than with the API.

Why it is worth closing rather than shrugging at: the field is agent-facing. An agent reading the tool schema sees a documented `effective_date`, receives `null` for every act, and has no way to tell "this act has no effective date" from "this server never fills that in". A dead field is worse than an absent one.

Options, in preference order:

1. Drop `effective_date` from the list-tool output shape, keeping it wherever act *detail* responses do carry the date. Cleanest, but changes the tool contract.
2. Populate it from a field the endpoints do return — `entryIntoForce` is present in the `acts/search` response and is plausibly what was meant. Needs checking against the detail endpoint's semantics before it is trusted.
3. Keep it and document it as reserved. Weakest option; leaves the confusing `null`.

Whichever is chosen, the recorded fixture now makes the outcome testable for real.

Issue #50, opened 2026-09-01
browse_acts test coverage gaps after the endpoint switch


numikel:
`browse_acts` moved from `acts/{publisher}/{year}` to `acts/search` in Klaster 8. The switch is guarded by three tripwires, but parts of the new behaviour are still unpinned.

- No test for the last partial page (an offset landing inside the final page).
- The page-shape test compares key names only, never values, so a field that changed meaning between the two endpoints would pass.
- The test page is a synthetic 20-item page built by duplicating one record, which cannot catch per-record differences.
- The main paging test is endpoint-agnostic, which concentrates the F30 guarantee into fewer tests than the file suggests.

---
Found during the whole-branch review of Klaster 8 (PR #43). Minor severity: not merge-blocking, filed so it is not lost.

PR #56, opened 2026-09-02
fix: close the deferred review findings from clusters 2–9 (4.2.0)

### LINKED ISSUE #14: Track MCP conformance support for protocol era 2026-07-28
## Summary

CI cannot run `@modelcontextprotocol/conformance` against the SDK protocol era advertised by `mcp[cli]==2.0.0`.

- Python SDK `LATEST_PROTOCOL_VERSION`: **`2026-07-28`**
- Conformance `@latest` accepted dated versions (as of 2026-08-10): `2025-03-26`, `2025-06-18`, `2025-11-25` (+ `draft` / `extension`)
- Probe error: `Unknown spec version: 2026-07-28`

## Design decision in this repo

Per migration design:

- Do **not** mask upstream lag by running conformance for an older protocol era.
- Direct STDIO / Streamable HTTP transport tests remain the mandatory protocol gate.
- When conformance gains `2026-07-28` support, CI should run the suite with `conformance-baseline.yml`.

## Tracking

- Example failing job: https://github.com/numikel/law-scrapper-mcp/actions/runs/31411065104
- Green skip job: https://github.com/numikel/law-scrapper-mcp/actions/runs/31411814338
- Upstream: https://github.com/modelcontextprotocol/conformance

Close when `@modelcontextprotocol/conformance` accepts `--spec-version 2026-07-28` and CI runs the suite without the skip warning.

numikel:
## Summary

Every open issue was re-verified against `v4.1.0` before anything was changed; three turned out to be already fixed or answered (#17, and items inside #19, #40, #38, #41), three are closed with a comment instead of code (#17, #44, #48), one is left for the owner (#36), and the remaining 22 are fixed here, one commit per issue. No MCP tool changes its name, parameter names or response shape.

**Truthful results**
- `effective_date` in `search_legal_acts`, `browse_acts` and `track_legal_changes` was always `null` — the code read a `dateEffect` key that neither list endpoint returns (the recorded fixture proves it). It now reads `entryIntoForce`, and the invented fixture that hid this is corrected (#52).
- `track_legal_changes` sends `limit`/`offset` upstream and reads `totalCount`, so a wide date range is stored as a `page`-scoped set with a real `corpus_count` instead of a truncated page labelled `complete` (#54).
- `search_legal_acts` validates `limit`, `offset` and `year` like every other listing tool and no longer forwards a negative offset to `api.sejm.gov.pl`; `limit` stays unclamped and its description now says why (#18, #19).

**Egress and download restraint**
- `LAW_MCP_API_MAX_SERVER_PAUSE` replaces the fixed 60 s constant; the clamp moved into the limiter so every `pause_for()` caller inherits it (#46, #47).
- Rate settings are bounded — `inf` was accepted and silently disabled pacing (#47); `backoff()` no longer overflows for large attempts (#27); the pacing deadline and the limiter share one clock, which made the deadline bound fail open when a limiter was injected (#45).
- HTML/PDF downloads are aborted mid-stream once they pass `LAW_MCP_DOC_STORE_MAX_SIZE_BYTES` (or earlier, from `Content-Length` on an unencoded body) instead of being materialised first; the budget is spent on decoded bytes, so compressed upstream responses are decoded exactly once (#19).
- `httpx`/`httpcore` loggers are held at `WARNING`, closing the F13 path where the full search URL reached INFO (#34).
- A cache entry for `acts/search` shared by several tools is now read within each caller's own TTL, so a call documented as 300 s is never served a 600 s entry (follow-up to #54, found in review).

**Hardening**
- `[::1].evil.com` no longer classifies as loopback, while a path suffix such as `http://[::1]:8080/` is ignored for every form; `StaticTokenVerifier` refuses an empty secret; `LAW_MCP_AUTH_REQUIRED_SCOPES` under `bearer` is rejected at startup rather than silently doing nothing (#38).
- `/health` is exempt from the per-client limiter only for loopback peers; JWKS discovery runs outside the verifier's lock (#39).
- `build_http_app()` derives auth from live settings (#41); `log_level` is a strict `Literal`, `shutdown_grace` an integer, `uvicorn` a declared dependency, plus the HALF_OPEN `/health` test and the `close()`-raises ordering test (#31).

**Document store**
- `DocumentStore.load` refuses oversized documents instead of truncating with a character slice against a byte budget (#32, #21); `end_pos == 0` is honoured and overlapping sections are rejected at load; two internal methods without callers are removed (#20); `search_in_act` builds only the requested page of match positions — measured 15.8 MB → 2.2 KB peak for a single-character query on a 1 MiB document (#16).

**Docker image did not build (found via the Glama indexer, not an issue)**
- The builder stage runs `uv sync --no-editable`, which builds the project wheel, and hatchling refuses to build without the declared `README.md` — which the Dockerfile never copied. Broken since the `readme` field was added on 2026-08-10; nothing in CI built the image. One `COPY` line fixes it, reproduced red/green by running the builder stage's own steps in a clean directory, and a test now derives the required files from `pyproject.toml` and checks the builder stage copies them.
- A new `docker` CI job builds the image and smoke-tests it the way its two consumers use it: `/health` over streamable-http with a throwaway bearer token, one `initialize` round-trip over STDIO (`LAW_MCP_TRANSPORT=stdio` overriding the image default, the shape registry runners use), and a non-root check.

**Tests and CI**
- CI runs the MCP conformance suite for real again: `@modelcontextprotocol/conformance` `0.2.0-alpha.11` is the first line that accepts the SDK's protocol era `2026-07-28`; the probe scenario switches to `tools-list` (`server-initialize` was removed at that era) and the limiter is disabled for the sweep, which otherwise produced 13 spurious 429 failures (#14, #39).
- Typed `outputSchema` pinning, `limit=0` upstream pin, browse last-partial-page and value assertions on the recording, shared fake clock, realistic clock magnitudes for the limiter tolerance, mid-wait pause and cancellation transitions, event-ordered arrival instead of wall-clock sleeps, shared client built from production settings, and a fully listed `TEST_SUITE_SUMMARY.md` (#49, #50, #55, #21, #40).

Ships as **4.2.0**: new operator setting, new download budget, and validation that turns previously silent behaviour into startup or tool errors (bearer scopes, `WARN`-style log levels, malformed `limit`/`offset`/`year`). Migration note for the one config-breaking change: remove `LAW_MCP_AUTH_REQUIRED_SCOPES` from bearer-mode deployments or switch to `oauth`.

Item 2 of #41 (`http/` shadowing the stdlib name) is a non-issue: it is a subpackage under absolute imports with no dynamic imports anywhere, so nothing to change. Planning artefacts stay out of the repository (`.ai/` is now gitignored alongside `docs/superpowers/`).

## How this was built

All 26 issues were verified read-only against `v4.1.0` first (file-and-line evidence per item, two live conformance probes, three claims reproduced by execution), and only the confirmed ones were implemented. Each code commit followed red → green → refactor: the failing test was written and run first, and its failure line is quoted in the commit body. The four independent areas (client/egress, search/changes, document store, auth/HTTP) were implemented in separate worktrees and integrated in order; the test-hygiene commits demonstrate that they catch the regression they guard by breaking the code locally first. A whole-branch review (comment below) found one Critical and four Important defects in the new code, all fixed red-first on the branch; the two findings it left standing (the IPv6 path suffix and the shared cache TTL) are fixed in the last two commits rather than deferred to issues.

## Test plan

- [x] Unit suite: 1359 passed (baseline on `main` was 1252), zero warnings
- [x] Integration suite (`-m integration`): 109 passed
- [x] Transport suites (STDIO + HTTP, `-m integration`): 9 passed
- [x] MCP conformance, run exactly as the CI step does: 27 checks pass, all 13 failures are in `conformance-baseline.yml` (baseline check passed)
- [x] Docker: builder stage reproduced outside Docker (no daemon in the sandbox), then the real `docker` CI job on this PR — build, HTTP smoke, STDIO smoke, non-root — passed
- [x] `scripts/check_release.py --version 4.2.0` — PASSED; `pip-audit` — no known vulnerabilities

Closes #14 #16 #18 #19 #20 #21 #27 #31 #32 #34 #38 #39 #40 #41 #45 #46 #47 #49 #50 #52 #54 #55


numikel:

Reviewed `83a7ab7..3c31191` (the branch as it stood: 78 files, 43 commits) at maximum effort through four lenses: streaming and egress correctness, error-contract integrity, parameter validation, and test isolation. Result: **1 Critical, 4 Important, 7 Minor**. Nine are fixed in the seven commits pushed since (`7e37154..f8f43fa`); four are left standing, with reasons. Every fix was written red first: the test was run against the previous code and its failure observed before the code changed.

**Critical — the new streaming download budget decoded every compressed body twice (`9bea1aa`).** `_receive_within` handed the materialised body back inside a response rebuilt from the original headers, `Content-Encoding` included — but `aiter_bytes()` had already applied the codec. httpx therefore ran gzip over plain text on `.text` and raised `DecodingError`; `classify_failure` left it non-retryable, `_request` turned it into `SejmApiError`, and `ActService._load_content`'s `except Exception` swallowed it, so `get_act_details(load_content=True)` returned `is_loaded=false` with no error and every later `read_act_content` failed. httpx advertises `Accept-Encoding: gzip, deflate, br, zstd` on every request, so this fires against any upstream that compresses. Reproduced locally as `SejmApiError: Błędne żądanie do API Sejmu: Error -3 while decompressing data: incorrect header check`. It is #19's own new code path, and nothing in the suite fed it a compressed body.

Two further defects in the same function shared one root cause — the budget being spent in wire bytes and decoded bytes interchangeably:

- **`Content-Length` answered for a quantity it cannot measure.** The pre-check compared wire bytes against a limit the document store enforces in decoded bytes, and reported the wire figure as `exact=True`. It now runs only for an unencoded body; a compressed one is settled one chunk later by the stream check, which counts what the store will actually hold.
- **The status gate was `response.is_error` where the unbudgeted path calls `raise_for_status()`.** A `300` (or any 3xx httpx does not follow, `has_redirect_location` being false) was not raised: its empty body was materialised and returned, so `get_act_html(..., max_bytes=…)` yielded `""` and stored it as the act's content. The gate is now `not response.is_success`, which is exactly what `raise_for_status()` rejects.

Four tests in `tests/unit/test_client/test_download_budget.py` cover this; three go red against the previous `_receive_within` (gzip round-trip, a wire length over budget for a body that fits, the 3xx refusal), the fourth pins that the stream check — not the declared length — is what answers for a compressed body.

**Important — `ResponseTooLargeError` broke the contract of the class it subclasses (`58356d2`).** The stated reason for the subclassing is that a refusal slipping through untranslated still lands in the same tool error category, but the subclass set neither `eli` nor `pdf_url`. Any `except ContentTooLargeError` reading them raised `AttributeError` on exactly that path, turning a `precondition` error into an `internal` one — the outcome the subclassing exists to prevent. They are now present and empty. The duplicated `exact` / `co najmniej` rendering moved into a shared `ContentTooLargeError._measured` in the same commit.

**Important — `search_legal_acts` still swallowed a malformed `year` (`af26be5`).** The #18 sweep made `limit` and `offset` loud but left `year` parsed under `contextlib.suppress` three lines above, on the one parameter whose loss changes the corpus rather than the page: `year="2o24"` dropped to `None`, the upstream `year` param was omitted, and the call searched all of DU while the agent was told nothing. It now raises the same Polish `ValueError` as its neighbours; the CHANGELOG entry was corrected to match.

**Minor, fixed:** the duplicated date on the 4.1.0 heading (`7e37154`); `DocumentStore.load` encoding the markdown to UTF-8 once for its own size check and again in `LoadedDocument.__post_init__` (`b7e4f6c`); and the `asgi_app` fixture calling `build_http_app()` without registering the two attributes it mutates on the shared `app` singleton, which leaked the last derivation into every later test in the session — the hazard `test_http_auth.bearer_app` already guards against (`65641f8`).

### What was left standing, and why

- **`build_http_app()` writing to the module-global `app` (#41).** The single derivation point is deliberate and documented; only the test-side leak above was closed.
- **JWKS discovery outside the verifier's lock (#39, D5).** Moving discovery out removed the serialisation without putting anything in its place: while an IdP hangs, `self._client` is never published, so every bearer-carrying request opens its own client against `/.well-known/openid-configuration` for the whole outage — the lockout traded for unbounded amplification against a provider already in trouble, which is the F26 shape. D5 calls the duplicate request "the cheaper wrong" and that is a defensible call; a single-flight task cached under the lock would give both properties, but with cancellation semantics to get right it is a design change, not a review fix.
- **`acts/search` cache entries are shared across tools while the TTLs differ.** Now that `track_legal_changes` sends `limit` upstream (#54), its key can be byte-identical to a `search_legal_acts` key: `get_json` keys on `f"json:{path}:{params or {}}"`, and for a publisher + date-range + limit call both services insert `publisher, dateFrom, dateTo, limit` in that order. Whichever writes first sets the TTL, so a call documented as 300 s can be served a 600 s search entry. The same collision between search and browse is already documented at `services/search_service.py:188` and embraced by D8 — `changes` joining it is new here, so it is flagged rather than changed. Worth a decision: a `cache_namespace` on `get_json`, or an explicit note that the longest TTL wins.
- **The tightened bracketed-IPv6 parser rejects `http://[::1]:8080/`.** The trailing slash leaves `_host_of` returning `[::1]:8080/`, `ip_address` raises, the entry classifies as remote, and startup under `auth_mode=none` refuses an allowlist entry that worked in 4.1.0. Fail-safe rather than fail-open, and loosening the suffix rule risks re-opening the `[::1].evil.com` hole the hunk closed — so it is recorded, not patched.

```
ruff check .                                All checks passed!               exit=0
ruff format --check .                       123 files already formatted      exit=0
mypy (project config, 54 source files)      Success: no issues found         exit=0
pytest (unit + in-memory integration)       1463 passed                      exit=0
pytest tests/integration/ -m integration    109 passed                       exit=0
scripts/check_release.py --version 4.2.0    PASSED                           exit=0
```

Five tests added (four on the download budget, one on `year`); no assertion was removed or weakened. The transport suites and the conformance sweep were left to CI, which runs both on this branch.

### Assessment

The Critical finding is a real, user-visible break introduced by this PR's own new code and would have shipped silently — a compressed act body producing `is_loaded=false` with no error. It is fixed with red-green coverage, as are the two accounting defects beside it. Of the four items left standing, two are deliberate design decisions restated here for the record, and two (`acts/search` TTL sharing, the IPv6 allowlist regression) are worth filing as follow-up issues. **Ready to merge** once CI is green on `f8f43fa`.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #59, opened 2026-09-06
fix: fail clearly on content loading errors

numikel:
## Summary

Signal content loading failures properly and cap error message length.

- Transient upstream failures (open circuit breaker, timeouts, 5xx) during `get_act_details(load_content=True)` now surface as `isError=true` instead of silently returning `is_loaded=false` with misleading hints.
- Permanent content absence (no HTML or PDF at all) returns success with new field `content_status="unavailable"` to distinguish from transient failures — the act exists, but has no readable text.
- Error messages from caller input or upstream responses are now length-bounded (configurable `LAW_MCP_ERROR_MESSAGE_MAX_CHARS`, default 500).
- All error messages end with a fixed one-sentence remediation hint ("Retry shortly", "Act not in registry", etc.).
- Regression tests freeze the v3.0.0 protocol fix: any domain exception surfaces as `isError=true`, never in a success envelope.

New `content_status` field on `get_act_details` output with three values: `not_requested`, `loaded`, `unavailable`. Hints for unavailable content now point to the PDF URL instead of to futile re-loading.

Changelog: v4.3.0 with BREAKING note — calls that previously returned success now fail with `isError=true`.

## Test plan

✓ Unit tests: 1412 passed
✓ Integration tests: 112 passed  
✓ ruff check: No issues
✓ ruff format: 124 files already formatted
✓ mypy: No issues

All acceptance criteria covered:
- Transient failures propagate as `isError=true`
- Permanent absence handled as success with `content_status="unavailable"`
- No placeholder documents in DocumentStore
- Hints avoid re-loading futile calls
- Error messages carry remediation guidance and respect length bounds
- Protocol-level regression frozen

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,8 +47,8 @@ readable text; `isError=true` means the upstream is unreachable and a retry may 
 ### Fixed
 
 - A failure to load content is no longer indistinguishable from an act without text.
-- `validation`, `not_found`, `precondition` and `unavailable` error messages are now
-  length-bounded.
+- `validation`, `not_found`, `precondition`, `content_too_large` and `unavailable` error messages
+  are now length-bounded.
 
 ## [4.2.0] - 2026-09-02
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,8 +36,8 @@ readable text; `isError=true` means the upstream is unreachable and a retry may 
 
 - `ActService._load_content` no longer swallows exceptions; only a permanent absence of readable
   text (no HTML/PDF URL, a 404 on the PDF fetch, or an empty extraction) is handled internally.
-- Hints for an act with no readable content point at the source PDF URL instead of tools that
-  would fail against it.
+- Hints for an act with no readable content point at the source PDF URL when the act has one,
+  instead of tools that would fail against it.
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,47 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [4.3.0] - 2026-09-06
+
+See [docs/changelogs/v4.3.0.md](docs/changelogs/v4.3.0.md) for details.
+
+Failed content loading now looks like a failure. Closes audit findings F33 and the remainder of
+F39, and freezes the protocol guarantees v3.0.0 introduced for F16.
+
+### BREAKING — content loading failures are protocol errors
+
+`get_act_details(eli=..., load_content=True)` used to return a successful response with
+`is_loaded=false` whenever loading failed. Such calls now fail with `isError=true`. A
+**successful** response with `content_status="unavailable"` means the act permanently has no
+readable text; `isError=true` means the upstream is unreachable and a retry may succeed.
+
+### Added
+
+- `content_status` field on `get_act_details` output (`not_requested`, `loaded`, `unavailable`),
+  part of the tool's `outputSchema` and `structuredContent`.
+- `LAW_MCP_ERROR_MESSAGE_MAX_CHARS` (default `500`, range 80-10000) caps the length of error
+  messages built from an exception's own text; truncation is announced in the message.
+- Every tool error message now ends with a fixed one-sentence remediation hint, one per error
+  category.
+
+### Changed
+
+- `ActService._load_content` no longer swallows exceptions; only a permanent absence of readable
+  text (no HTML/PDF URL, a 404 on the PDF fetch, or an empty extraction) is handled internally.
+- Hints for an act with no readable content point at the source PDF URL instead of tools that
+  would fail against it.
+
+### Removed
+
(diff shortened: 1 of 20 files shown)


numikel:

Reviewed `bc69f04..8a1376e` against the Klaster 10 spec and plan through two independent lenses: plan alignment / architecture / cross-task integration / test quality, and security (information disclosure, truncation correctness, egress). Result: **0 Critical, 2 Important**. Both are fixed in `32baaa3` (plus a comment-only follow-up `6f8b272`, see below) and the fixes were confirmed by a second pass of each reviewer. All 18 acceptance criteria of spec §5 were checked one by one and each has a behaviour-asserting test or a gate run; every D5 failure path was traced through the code individually.

**A 404 on `text.html` was not the permanent absence D5 says it is.** Only the PDF branch of `_load_content` mapped `ActNotFoundError` to `ContentNotAvailableError`; an act whose metadata promised HTML that then 404ed failed the whole call as `not_found` with the guidance "this resource does not exist in the registry" — for an act whose metadata had just been fetched successfully in the same call. The gap was documented as "known, deferred" in the changelog, but never approved the way the equally sized D9 question was. The HTML branch now mirrors the PDF branch; `test_missing_html_is_a_documented_absence` mirrors the PDF test one for one, and the "known gap" paragraph is gone from the changelog, README and docstring.

**Removing the placeholder documents (D3) removed an accidental negative cache.** Before this branch, a textless act got a placeholder in `DocumentStore` with `is_loaded=true`, so every later `load_content=true` on it was a no-op upstream. After D3, nothing was written, so every repeat spent one or two requests on `api.sejm.gov.pl` for an answer that cannot change before the metadata it was derived from expires. The letter of O1 held (no retries inside `_load_content`), its spirit did not. `ActService` now keeps a bounded `TTLCache` of ELIs the source has declared textless, with `cache_details_ttl` as its TTL — the same value as the metadata whose `has_html`/`has_pdf` flags route the fetch. `test_permanent_absence_is_remembered` asserts one upstream request across two calls. The security reviewer confirmed the key space is bounded to registry-confirmed ELIs (an invented one 404s at the metadata call and never reaches the cache) and that the HTML-404 mapping is gated on a literal status 404, not a heuristic.

### What held up under scrutiny

- **The transient/permanent boundary is exactly D5's table** — open breaker, timeout, 5xx, converter exception all propagate to `isError=true`; neither format, empty extraction and (now) 404 on either endpoint end as `content_status="unavailable"`; `ContentTooLargeError` keeps its O2 behaviour.
- **`content_status` is derived in one place from `is_loaded`** (R2), and the shipped code improves on the plan's own pseudocode: a metadata-only call on an already-loaded act reports `loaded` rather than `not_requested`. The negative cache sits strictly downstream of that derivation and cannot desynchronise the two fields.
- **F16 and F39 are frozen by real tests** through the in-memory `mcp.Client`, asserting `isError=true`, no `error` key in a success body, and the absence of a 40 kB upstream body from the message — not vacuous shape checks.
- **Truncation is safe by construction**: `error_message_max_chars` is bounded `ge=80`, so the slice never goes negative; guidance is appended after the cut, so it always survives; a bad env value fails startup loudly.
- **File scope matches spec §6 exactly**; no test was weakened or deleted; the five version locations agree at 4.3.0; English-only convention held in code, tests and changelog.

Run locally from the branch worktree at `32baaa3`; the comment-only follow-up re-ran `ruff check`, `ruff format --check` and the services test package.

```
ruff format --check src tests               124 files already formatted     exit=0
mypy src/law_scrapper_mcp/                  No issues found, 54 files       exit=0
pytest tests/unit/                          1414 passed                     exit=0
pytest tests/integration/ -m integration    112 passed                      exit=0
pytest transport (stdio + http)               9 passed                      exit=0
scripts/check_release.py --version 4.3.0    PASSED                          exit=0
pip-audit (project venv)                    No known vulnerabilities found  exit=0
```

Unit tests 1412 → 1414. CI on `32baaa3`: all ten jobs green.

Not merge-blocking, filed so they are not lost: #60 (`content_too_large` messages can lose their PDF URL under a `LAW_MCP_ERROR_MESSAGE_MAX_CHARS` near the 80-char floor; untested), #61 (`ApiUnavailableError` echoes the raw `httpx` exception into the caller message and the unredacted ERROR log — pre-existing, outside this cluster's file scope; the one path where F39 does not hold by construction). Two further minors from the first pass (README silent on the HTML-404 gap, docstring/changelog drift about it) became moot with the fix; a fourth (the negative cache shares a TTL *value* with the metadata cache, not a clock, so a verdict can outlive its snapshot by up to one extra window) is fixed in the comment-only commit. Cosmetic grapheme splitting at the truncation boundary was noted and deliberately not filed.

### Assessment

**Ready to merge.** D1–D9 hold, the plan's eight tasks are all present with their acceptance criteria covered, no commit undoes an earlier one, and outbound traffic went down rather than up (no request for an act with neither format, no repeat fetch for a remembered absence). One design addition beyond the spec — the negative cache — is recorded in the changelog and README; if the maintainer prefers a different TTL source for it, that is a one-line change.

